PR #26, opened 2015-12-02
Fix SSL cert error on data update

mbadams5:
I was getting an SSL cert failure on a fresh Windows 8 VM, causing the data update to fail.

I tested this on Windows 8 and my Mac.


asparke2:
@mbadams5 does this fix @piljaeim 's issue?  I don't have a good way to test.

PR #37, opened 2016-02-10
Update Large Office/Warehouse

mbadams5:
There were a few nagging issues with these two building types. One was using a FluidCooler for the heat pump loop, which is still not fully fixed. Currently it uses a CoolingTower:TwoSpeed instead and outputs match closely. 

There were various issues in the spreadsheet for construction sets, curves, and such which were updated to make simulations run to completion. 

The Pre-1980 and Post-1980 Warehouse models have very high cooling and fan energy which still need to be investigated further. Other than that, all vintages for Warehouse and Large Office are very close to the original models now.


amirroth:
@mbadams5, many of the LargeOffice models are showing Severe warmup convergence errors, e.g., 
`** Severe  ** CheckWarmupConvergence: Loads Initialization, Zone="PERIMETER_MID_ZN_1 ZN" did not converge after 25 warmup days.` Just an annoyance? Or a serious problem?

mbadams5:
Usually just an annoyance, but i'll check to make sure it isn't anything more. The Mid zones have a multiplier on them so I don't know if that is affecting this any or not.

amirroth:
Thanks. If these are annoyances, shouldn't they be warnings instead of
severe errors?

mbadams5:
Well if zone does not converge in the maximum number of warmup days, EnergyPlus always treats it as a severe error. I'll have to look back at the old legacy IDFs and see if they have convergence issues or not.

mbadams5:
@amirroth I ran the legacy prototype IDF `ASHRAE90.1_OfficeLarge_STD2013_Chicago.idf` with OHare weather file and it has two zones that fail warmup convergence. So it might be something inherent to how these prototype buildings are constructed to begin with and that is, of course, following through to the OS generated buildings.

PR #68, opened 2016-05-08
Fix plant loop configurations

lincolnharmer:
Removes the dependency on the prototype methods for CHW plant creation. This allows for the creation of CHW and CW loops more in line with App. G.


asparke2:
This is covered in the baseline_automation_90.1_fixes branch I just merged

PR #71, opened 2016-05-12
Correct classification of WSHP, and check for HX connected to boiler on secondary PL

jmarrec:
Added correct classification of Water Source Heat Pumps as being potentially Hybrid if serves by hot water loop. And added check for plant loop with heat exchanger


asparke2:
Looks good except for WSHPs being fueled by the fuels on the loop as well.

Issue #73, opened 2016-05-15
non-integrated air-side economizers not being specified properly


lincolnharmer:
The Lockout Type for the Controller:OutdoorAir object is being set to "NoLockout" for non-integrated economizers [here](https://github.com/NREL/openstudio-standards/blob/master/openstudio-standards/lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb#L1025). This means that when the unit is in heating mode the outdoor air damper is not at its minimum position like it should be for correct economizer control. This results in excessive heating energy use. Setting this line to "LockoutWithHeating" should fix the problem.

lincolnharmer:
Just made a commit to the [fix non integrated airside economizers branch](https://github.com/NREL/openstudio-standards/tree/fix-non-integrated-air-side-economizer) to fix this

asparke2:
I'll talk to the PNNL folks about this one.

PR #74, opened 2016-05-15
setLockoutType to 'LockoutWith Heating' instead of 'NoLockout'

lincolnharmer:
addresses issue #73


asparke2:
This doesn't jive with the DOE prototype buildings, not going to merge

Issue #77, opened 2016-05-18
Economizers aren't being enabled for PSZ-AC


jmarrec:
Need to set the economizer type in Standards.AirLoopHVAC. The economizer type is determined but never set (the temperature/enthalpy limits are).

Around https://github.com/NREL/openstudio-standards/blob/master/openstudio-standards/lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb#L1231

Add

    # Set the Economizer Type
     oa_control.setEconomizerControlType(economizer_type)

    

I'll do a PR once I'm done working on my current project (for which I've modified a lot of code to suit my own needs)

asparke2:
Fixed by https://github.com/NREL/openstudio-standards/pull/95

PR #81, opened 2016-05-19
Set economizer properly (see #77)

jmarrec:
https://github.com/NREL/openstudio-standards/issues/77


mdahlhausen:
Economizer settings were added to openstudio-standards a while ago.  I copied the pressure drop adjustment code into #808.

Issue #96, opened 2016-07-22
PRM: gracefully handle Service Hot Water Loops


jmarrec:
This is potentially going to be tricky, but we have to find a way to gracefully handle Service Hot Water Loops.

**Right now the current logic is:**
- a plant loop that serves a water use connections isn't deleted [Standards.Model.create_performance_rating_method_baseline_building](https://github.com/NREL/openstudio-standards/blob/master/openstudio-standards/lib/openstudio-standards/standards/Standards.Model.rb#L156) calls [Standards.Model.remove_performance_rating_method_hvac](https://github.com/NREL/openstudio-standards/blob/master/openstudio-standards/lib/openstudio-standards/standards/Standards.Model.rb#L3720)
- Get all WaterHeaters in model and change fuel type if standard is 90.1-2013, do nothing otherwise.

**Couple of Notes:**
- The WaterHeater fuel type is one thing, but there's a ton more to do, per G3.1.11 (especially b. and e.)
- eg G3.1.11.e: if a "Where a combined system has been specified to meet both space heating and
  service water heating loads, the baseline building system shall use separate systems meeting the minimum efficiency requirements applicable to each system individually.". This is quite common in my experience to have a space heating boiler serve a storage tank through an internal or external HX. In the baseline I'd model a `WaterHeater:Mixed` with a minimum efficiency requirement (typically 80%), having the same storage capacity as the proposed design, and a  burner heating capacity similar to the proposed system (this is tricky, the space heating boiler really is sized for the space heating load + DHW. I know some reviewers that force you to use the proposed space heating boiler capacity in the baseline waterheater, which to me is completely ridiculous, but I digress).
- I just ran the PRM on a model that had a Service Water Heating loop with a single branch on the supply consisting of a FluidToFluid:HX and a WaterHeater:Mixed used as a storage tank only (no burner capacity). The resulting thing is en E+ crash because the FluidToFluid:HX is no longer connected on the supply side. The waterheater:mixed still has zero burner capacity anyways.
- G.3.11.b is especially fundamental too: in any case you should model the baseline per minimum efficiency requirements, and if proposed heating source is electricity, then electric resistance is used (particularly important in case proposed is a Heat pump water heater)

**Suggestion:**

I suggest trying to asses how the SHW load is being handled: what fuel type is used, and is this a combination of space heating and SHW. Find the storage capacity of any storage medium, and find the heating capacity of the equipment serving the SHW load. 

[...]

PR #97, opened 2016-07-22
Correctly classify WSHP and handle WaterHeaters 

jmarrec:
Correctly classify WSHP (see #71), and handle WaterHeaters (Mixed and…Stratified) correctly by checking it it's a storage tank only, and whether it's connected on the source side to a plant loop (related #96)


mdahlhausen:
@asparke2 this is really old.  I want to close it.  Please see what we should use from this pull request and make a new pull request with that content, then close this one.

mdahlhausen:
@asparke2 thinks this code was merged previously

Issue #108, opened 2016-10-03
Large office shorter than multiplier would suggest


DavidGoldwasser:
Seems the multiplier is fine, but the mid and top floor is just positioned lower than it should be. Minor issue, but would be nice to fix at some point.
https://unmethours.com/question/18033/why-is-large-office-prototype-building-so-short/

DavidGoldwasser:
Here is geometry for the large_office_reference.osm

![screenshot 2016-11-02 11 54 34](https://cloud.githubusercontent.com/assets/4857055/19941092/6b434aa2-a0f4-11e6-9390-bb32533b4fea.png)

large_office.osm and large_office_2010.osm seem to share the same geometry at first glance (at least height profiles for different spaces).

![screenshot 2016-11-02 12 00 13](https://cloud.githubusercontent.com/assets/4857055/19941139/952703e0-a0f4-11e6-8664-05ee27eea065.png)

@asparke2 I can fix the mid and top floor positions to match large_office.osm without any problem. I did notice the basement height varies as well, I assume that was an intentional decision in support of the data center or for other reasons?. Charnged form 2.2394m to 2.4390m

asparke2:
Yeah it was probably for the data center addition

DavidGoldwasser:
Here is fixed geometry

![screenshot 2016-11-02 12 23 24](https://cloud.githubusercontent.com/assets/4857055/19941899/7fdc173e-a0f7-11e6-91bd-99b76c958b5a.png)

Issue #119, opened 2016-10-25
Geometry for Midrise Apartment is inside-out


asparke2:
The geometry for the Midrise Apartment is inside-out:

This geometry was copied directly from the DOE Prototype Building IDF file, which is also inside-out:
![image](https://cloud.githubusercontent.com/assets/2009290/19701187/85d9845a-9ab7-11e6-8686-1ffbbdd89596.png)

The DOE Reference Buildings are correct:
![image](https://cloud.githubusercontent.com/assets/2009290/19701189/8909ccca-9ab7-11e6-9e15-c59ee7ac9652.png)

This did not show up in the comparison tests because both the OSM and the original IDF were wrong in the same way.

eringold:
This is some Treehouse of Horror-type stuff. 
![tumblr_mcc04qq9vv1qzztl4](https://cloud.githubusercontent.com/assets/10799524/19702270/05f80464-9abc-11e6-8575-6d7cab435d0f.jpg)

DavidGoldwasser:
@asparke2 I noticed that the first and second floors are not matched, but since the second floor has a multiplier of 2, seems like adiabatic is correct since the heat transfer would be asymmetrical. 

![screenshot 2016-11-02 10 11 52](https://cloud.githubusercontent.com/assets/4857055/19937230/f8c3ece2-a0e5-11e6-82aa-6fddf5efacbb.png)

If someone runs this model through surface matching for any reason it will automatically match the surfaces. Could be wise to raise the mid floor up so it is floating between the bottom and top floor, which is more typical configuration when zone multiplier is used as a 'floor multiplier'.

Although as devils advocate, idea of surface matching messing things up wouldn't be unique to this prototype building type, so maybe we should leave it alone? Another example is large hotel where multipliers are not just used as floor multipliers, but as a room multiplier. Could just have guideline 'Don't run surface matching on prototype geometry'. Hotel doesn't have an easy work around for this, unless surface matching is setup to ignore adiabatic surfaces.

DavidGoldwasser:
For reference here is before and after end use breakdown. Total EUI was roughly 43.5 with old (reversed wall model) and 43.96 in fixed model. This run was done with hand fixed model. results when fixed geometry used in measure may be different if it affects sizing run or other characteristics. Heating end uses went down in the fixed model while cooling and fan consumption went up. Some zone conditions were affected as well.

[...]

PR #123, opened 2016-11-02
Reversing wall and sub-surfaces re issue #119

DavidGoldwasser:
re Issue #119


DavidGoldwasser:
Should see a bunch of changes like this where surface and sub-surface vertices were reversed.

screenshot 2016-11-02 11 14 38

DavidGoldwasser:
@asparke2 want me to fix issue #108 with this pull request as well?

asparke2:
@DavidGoldwasser sure that would be good.  Please check the original Prototype and Reference IDFs to make sure it matches them.  If they also appear to have issues, we need to tell PNNL.

DavidGoldwasser:
@asparke2 based on the details in UH post it seems like the Reference model correctly was 47.6m while the prototype model was just 37.5m. Sounds like @jmarrec reached out to PNNL but we can follow up on that. I'll fix the model in openstudio-standards. I'll check all three geometry models for large_office.
https://unmethours.com/question/18033/why-is-large-office-prototype-building-so-short/?answer=18036#post-id-18036

PR #229, opened 2017-04-21
Miscellaneous code fixes

eringold:
- fix units for `oa_for_volume`
- fix leaving condenser water temp determination


jmarrec:
Yay  First-time contributor! Good catch and good job on the PR

PR #258, opened 2017-07-13
Bricr lbnl

macumber:


asparke2:
This is old, will merge the updated one when able

PR #308, opened 2018-03-12
align rubocop with openstudio resources

nllong:
Align rubocop with the contents in openstudio-resources. 

@asparke2 -- I would like just reference the version on OpenStudio-resources like [this](https://github.com/NREL/OpenStudio-server/blob/develop/.rubocop.yml). The only draw back is that at NREL you have to be on devvpn so it can download the file over https.


asparke2:
changing to a reference works for me, dev vpn shouldn't be an issue

asparke2:
also, can you change the target of this PR to release/0.2.0?  I will run rubocop on that branch as part of release process.

nllong:
I updated the base branch and made it a reference.

PR #312, opened 2018-04-05
Nrcan ckirney

ckirney:
Refactored necb_spacetype_system_selection method in NECB2011 class (in hvac_systems.rb file) to use necb_hvac_system_selection_type.json for selecting hvac system types rather than using case statement.  Adjusted necb_hvac_system_selection_type.json to fix some incorrect hvac system selections, include a min_cooling_capacity_kw key to make refactor of necb_spacetype_system_selection more sensible, and added reference to table in NECB 2011.  Adding sched_create utility to read schedules from NECB 2015 tables spreadsheet and put them in a JSON file.


ckirney:
Canceled CircleCI tests were for a file which I changed and then reverted back to the original version.  Tests 1315 and 1316 are those relevant to this commit.  Currently running test 1319 which is for a very minor change to necb_hvac_system_selection_type.json.  Will very likely pass that test but you can wait for it to complete first.

PR #317, opened 2018-04-18
Nrcan ckirney

ckirney:
Modified schedules.json and necb_hvac_system_selection_type.json to incorporate template key which identifies which NECB version the file was created from.  Added schedulesNECB2015.json which is currently contains NECB2011 data and is a placeholder.  It will be replaced with real NECB 2015 schedules once I have confirmed that BTAP can handle the schedules appropriately.  Added a utility which add a template key and appropriate NECB version value to necb json files.  Adjusted sched_create utility to incorporate template key when creating a schedule json.


ckirney:
Also, I changed NECB_2011.rb to incorporate support for more schedule types.

PR #368, opened 2018-06-19
added multipliers for space and subsurface area calculation

padmassun:
https://github.com/canmet-energy/btap_tasks/issues/79


padmassun:
Branch can be deleted after merge

PR #371, opened 2018-06-22
add sensible and latent fractions to DEER SHW space types where missing

mdahlhausen:
needed for UrbanOpt runs


mdahlhausen:
already merged

PR #383, opened 2018-06-28
SmallOfficeDetailed

piljaeim:
Revised prototype small office building model with detailed space types. The model was developed based on older version of the code, so it may need to be update per recent changes.


asparke2:
Small Office Detailed merged by #562

PR #384, opened 2018-06-28
 Add coil_cooling_water as part of the economizer_compliance

padmassun:


padmassun:
https://github.com/canmet-energy/btap_tasks/issues/99

PR #428, opened 2018-08-15
[Nrcan Task 127] For local testing, modify test such that circelci_tests.txt file does not get modified

padmassun:


padmassun:
Moved to  https://github.com/NREL/openstudio-standards/pull/432

PR #506, opened 2018-10-03
set optional double to value

mdahlhausen:


mdahlhausen:
@asparke2 this error was caused by an [OpenStudio commit](https://github.com/NREL/OpenStudio/issues/2625) in the AirTerminalSingleDuctVAVReheat object changing the .fixedMinimumAirflowRate method to return an optional double instead of a double.

mdahlhausen:
@asparke2 this is failing because the ci tests are using OpenStudio v2.6 series instead of v2.7

PR #508, opened 2018-10-08
handle terminal method returns as optional double instead of double

mdahlhausen:
AirTerminalSingleDuctVAVReheat method constantMinimumAirFlowFraction returns optional double instead of double in OS v2.7

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: ~/openstudio-standards
     docker:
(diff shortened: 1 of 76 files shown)


asparke2:
This merge fixes a few issues that were caused by updates in OpenStudio 2.7.0, including:
1. Changed return types of some OS model methods in VAV terminals
2. EnergyPlus 8.9 update required addition of new fields to ddy files
3. openstudio-cli changed how external gems were loaded, which necessitated changing how openstudio-standards calls the cli internally, which mainly happens for sizing runs

PR #513, opened 2018-10-12
Parallelize export of OpenStudio libraries

jmarrec:
I modified `export_OpenStudio_libraries.rb` to use [Parallel](https://github.com/grosser/parallel) because it's taking such a long time and each iteration of the for loop is really independent.

```ruby
gem install parallel
bundle exec rake library:export
```

I also caught a few weird things in the script. If it can't find the template in question, it still tries to do it which will fail. And currently there was a line that would basically only make it run for 90.1.2004

### DIFF
--- a/data/standards/export_OpenStudio_libraries.rb
+++ b/data/standards/export_OpenStudio_libraries.rb
@@ -13,18 +13,18 @@ require 'etc'
 # gem install parallel
 require 'parallel'
 
-# Environment variables
-if ENV['N'].nil?
-  # Number of parallel runs caps to nproc - 1
-  $nproc = [1, Etc.nprocessors - 1].max
-  puts "Defaulted Nproc to #{$nproc}"
-else
-  $nproc = ENV['N'].to_i
-  puts "Using environment Nproc=#{$nproc}"
-end
-
-
 def export_openstudio_libraries
+
+  # Environment variables
+  if ENV['N'].nil?
+    # Number of parallel runs caps to nproc - 1
+    $nproc = [1, Etc.nprocessors - 1].max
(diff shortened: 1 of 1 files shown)


jmarrec:
Note that it defaults to `nproc-1`, with in my case is 15 threads. But that still "only" loads my CPUs to about 40%. I didn't test increasing it N to double that amount, but mind that it might cause some RAM usage problems on some smaller machines (for info it takes 5GB with 15 threads).

phylroy:
Hey J,

I'm wondering why you are using the extra gem etc instead of using the parallel gem's method to determine the number of processors available on the system. 

Below is an example of how we run the tests locally before we send to circleci. 

https://github.com/NREL/openstudio-standards/blob/nrcan/test/test_run_all_test_locally.rb#L9

jmarrec:
@phylroy mostly because I never heard of parralel before yesterday so I didn’t know it had such a method built in.
Might as well use that one yeah!

jmarrec:
Really just out of curiosity: [here](
https://github.com/grosser/parallel/blob/fe6f49c5599bd9fa3ea3a8a5d883602f6128934d/lib/parallel/processor_count.rb#L21) is where Parallel.processor_count is defined.

I thought it might use Etc, which it does by default. Unless it doesn’t exist then it goes the extra mile.

jmarrec:
Running the script took 19 hours, even parallelized... Has that been your experience too?

Also DEER 2014, 2015 and 2017, and DEER_pre_1975, and the two DOE_REF  (DOE_Ref_1980_2004 and DOE Ref Pr 1980) didn't get generated, not sure why/if normal.

PR #515, opened 2018-10-19
zone splitting bug fix

mdahlhausen:


lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -1211,11 +1211,14 @@ def model_eliminate_outlier_zones(model, array_of_zones, key_to_inspect, toleran
...
-    biggest_delta_i = nil
+    biggest_delta_i = 0 # array at first item in case delta is 0
     biggest_delta = 0.0
     worst = nil
     array_of_zones.each_with_index do |zn, i|
       val = zn[key_to_inspect]
+      if worst.nil? # array at first item in case delta is 0
+        worst = val
+      end
       delta = (val - avg).abs
```
macumber: Do you still need to handle case where total_area is 0 on line 1207?

macumber: If total_area is zero then avg will be infinity and delta = (val - avg).abs is not well behaved

mdahlhausen: added a guard clause

PR #544, opened 2018-11-20
Reinstated test_necb_shw and revised method to check if spaces are heated, cooled, or plenums 

ckirney:


ckirney:
Now checking if spaces are heated or cooled using the space type data in the standards_data table.  If a heating or cooling schedule is included then the spaces are considered conditioned.  Also, now determine if a space is a plenum by looking at the space type data in the standards_data table.  If no NECB hvac type is mentioned for the space then it is considered a plenum.  Finally, modified test_necb_shw.rb and the expected results from that test to account for the changes to how mechanical rooms are identified.  Reinstated the test_necb_shw.rb in the circleci tests.

PR #547, opened 2018-11-28
Nrcan 157

phylroy:


IainMacdonald:
Phylroy,

Just found this in my junk… From what we discussed a couple of weeks ago I’m thinking that this is the notification we need to ‘update’ our local files. Is that correct?

I’m also wondering if the message was filtered for Edward and Marianna. I get most but the occasional one gets filtered out. Weird.

Iain

PR #548, opened 2018-11-30
Fix/update deer curves

mdahlhausen:
- sets temperature limits for deer curves
- fills in missing curve lookups for dx equipment with NECB defaults
- fixes temperature limit defaults for defrost EIR curves in prototype library


mdahlhausen:
curve edits included in #605

PR #555, opened 2018-12-05
mandate off-hour controls for air loops per 90.1 section 6.4.3.3

mdahlhausen:


mdahlhausen:
Summary of decision:

In openstudio-standards HVAC air system schedules are set here:
https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb#L145-L159

The ‘air_loop_hvac_enable_unoccupied_fan_shutoff’ sets the air loop availability schedule with a default 5% occupancy threshold, and the ‘air_loop_hvac_add_motorized_oa_damper’ method sets the minimum outdoor air schedule with a default 15% occupancy threshold.

The test for whether to apply ‘air_loop_hvac_enable_unoccupied_fan_shutoff’ is the ‘air_loop_hvac_unoccupied_fan_shutoff_required?’ method, which tests against ASHRAE 90.1 6.4.3.4.5: systems greater than 0.75 HP must turn off when unoccupied.
This is listed in 90.1 under ventilation fan control not system level control, which is specified in section 6.4.3.3.1 Automatic Shutdown, essentially requiring shutting the system off per a schedule or after 30 minutes of no occupancy.

Most systems pass the 0.75 HP threshold, and the system gets shutoff regardless of whether it has a motorized OA damper or not.  However, many single zone systems have fans smaller than 0.75 HP and often don’t need to meet the motorized OA damper requirements either.  These systems are then defaulted to always on, with an always on minimum ventilation schedule, causing a lot of excess heating/cooling/fan energy during unoccupied periods.

Functionally what this means is that our LA ComStock and DEER models with PSZ-AC or PSZ-HP systems serving small spaces have much higher fan energy use than the CEUS comparison set.

In reality, even the DOE prototype models have off hour controls, and this is required in ASHRAE 90.1 section 6.4.3.3.

The chosen fix in this pull request is to default set the availability schedule to mirror the occupancy schedule for all HVAC air loops, removing the 0.75 HP fan threshold requirement.

Issue #575, opened 2019-01-25
space_apply_infiltration_rate(space) results in no infiltration for DOE Ref Pre-1980 and DOE Ref 1980-2004 


DavidGoldwasser:
They don't have a template specific logic rend relies on the default that uses 0
https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/standards/Standards.Space.rb#L1294-L1297

I think we should add template specific code for these vintages based on what was used for the reference buildings. I'm also not a big fan of the default of 0, I'd almost rather have it fail then not add infiltration. This doesn't affect the DOE prototype measures, but it does impact create_typical_building_from_model.

cc: @macumber @rHorsey @mdahlhausen @asparke2 

I'm assigning myself, but if someone beats me to it that is fine, but would be nice to get in 2.8.0

DavidGoldwasser:
Information about ref buildings from section 5.3.2 Infiltration 
in U.S. Department of Energy Commercial Reference Building Models of the National Building Stock PDF
https://www.nrel.gov/docs/fy11osti/46861.pdf

> For existing buildings, we turn to measured airtightness data. Persily (1998) collected data from
139 buildings that showed a wide scatter in airtightness and little correlation between building
age and airtightness. The average leakage rate across all building types and locations was 1.5
cfm/ft2 at 0.3 in. w.c. (7.5 L/s/m2 at 75 Pa) of above-grade envelope area. This number is used for the pre-1980 and post-1980 reference building models.

mdahlhausen:
[Initial commit to fix this](https://github.com/NREL/openstudio-standards/tree/fix/doe_ref_infiltration_default).  We may need additional logic to override the defaults when making the DOE ref prototypes to use 1.5 instead of 1.8 cfm/ft2 exterior surface area at 75 Pa.

mdahlhausen:
Fixed with pull #576

PR #576, opened 2019-01-29
Fix/doe ref infiltration default

mdahlhausen:
Three changes beyond comments:
- sets the default infiltration rate to the PNNL modeling guideline value of 1.8 cfm/ft2 exterior *surface* area at 75 Pa.
- Changes the space_apply_infiltration_rate method to use values as if they are correctly exterior *surface* area, not exterior *wall* area as was being done in the method.
- Updates the default DEER space infiltration rate of 0.038 cfm/ft2 exterior *wall* area at natural pressure assumed in the DEER BDL rules to match a value given in cfm/ft2 exterior *surface* area at 75 Pa assumed in the code.

Should fix Issue #575


lymereJ:
@mdahlhausen, the proposed changes make the 90.1 prototypes generated by the OpenStudio-Standards gem diverge from the Commercial Building Prototypes. In the Commercial Building Prototypes infiltration is modeled using the guidelines in [Gowri et al](http://ibpsa-boston.com/minutes/IBPSAboston-presentations_2011-09-16_Gowri.pdf) (see section 6). Infiltration flow is provided as a function of above grade exterior wall only. Most prototypes use `Flow/ExteriorArea` as design flow rate calculation method so the infiltration flow rate per above grade exterior wall area has to be converted to flow per exterior surface area. The proposed changes keep the same design flow rate calculation method (`Flow/ExteriorArea`) but removed the flow adjustment. For example, for the high rise apartment prototype generated by the OpenStudio-Standards gem, for the “T S1 Apartment” zone, the flow per exterior surface area used to be 1.6E-4 (similar to what is in the Commercial Building Prototypes) but is now 5.6E-4 cfm/ft2 of exterior surface area. Either the design flow rate calculation should be updated to `Flow/ExteriorWallArea`, or the flow adjustment should be brought back. Let me know if I'm missing something, thanks.

mdahlhausen:
@lymereJ, @asparke2 I think there was a typo in the "*Infiltration Modeling Guidelines for Commercial Building Energy Analysis*" report you linked to above, thus the change.  The recommendations in part 6 state:

*"Following the above methodology, the EnergyPlus input design infiltration (Idesign) was calculated as 0.2016 cfm/ft² (0.001024 m3/s/ m2) of above grade exterior wall surface area, equivalent to the base infiltration rate of 1.8 cfm/ ft² (0.00915 m3/s/ m2) of above-grade envelope surface area at 0.3 in. w.c. (75 Pa)."*

However, 1.8 cfm/ ft² (0.00915 m3/s/ m2) of above-grade envelope surface area (the Envelope Subcommittee recommendation in the first paragraph of section 2 and the sum in section 2 Table 1) correlates to a design infiltration rate of 0.2016 cfm/ft² (0.001024 m3/s/ m2) of above-grade **envelope** surface area per ASHRAE 90.1-2010 PRM Reference Manual section 3.2.5, which also matches the last paragraph of section 4:

*"The calculated value for Idesign on a per square foot basis for each benchmark building is then 0.2016 cfm/ft2 for all above ground envelope area. This is equal to 0.001024 m3/s-m2."*

Given that the 1.8 number is calculated for total above-grade envelope surface area, I'm inclined to think that the 0.2016 number is also above-grade envelope surface area and not above grade exterior wall surface area as mentioned in the conclusions section.  This would make more sense, as a per wall-area value would imply more infiltration, and the default value is already 20% higher than the mean value of 1.5 cfm/ ft² of above-grade envelope surface area found for existing buildings in A. Persily, *"Airtightness of Commercial and Institutional Buildings: Blowing Holes in the Myth of Tight Buildings"*, 1998 that was used by the Envelope Subcommitte and PNNL as a reference.

Furthermore, [the Prototype Model .idfs](https://www.energycodes.gov/development/commercial/prototype_models) give infiltration at 0.001024 m3/s/ m2 per exterior area.

```
  ZoneInfiltration:DesignFlowRate,
    Perimeter_top_ZN_2_Infiltration,   !- Name
    Perimeter_top_ZN_2,       !- Zone or ZoneList Name
    INFIL_SCH_PNNL,           !- Schedule Name
    Flow/ExteriorArea,        !- Design Flow Rate Calculation Method
    0,                        !- Design Flow Rate {m3/s}
    ,                         !- Flow per Zone Floor Area {m3/s-m2}
    0.001024128,                !- Flow per Exterior Surface Area {m3/s-m2}
    ,                        !- Air Changes per Hour {1/hr}
    0,                       !- Constant Term Coefficient
    0.0000,                  !- Temperature Term Coefficient
    0.2240,                  !- Velocity Term Coefficient
    0.0000;                  !- Velocity Squared Term Coefficient
```

So, this change to openstudio-standards re-aligns to the original prototype .idfs and the apparent original intent of the envelope subcommittee.

lymereJ:
@mdahlhausen, @asparke2 -

We contacted with the author of Gowri et al (2009) and confirmed the infiltration rate of 1.8 cfm/sf2 of above-grade envelope surface area, at testing condition (75Pa), was determined by ASHRAE Standard 90.1 Envelope Subcommittee when they tried to evaluate the relative impacts of two continuous area barrier requirements, Addenda AM and BF to 90.1-2007. The committee reviewed Persily’s paper but chose to not use the mean value of the source data.

This value is converted for input to EnergyPlus as an average infiltration rate which is function of the wind speed using the average of the positive surface pressure coefficients around all sides of the building, it does not include the roof. The paper specifies that higher air speeds are assumed across the roof surface and hence are not expected to increase infiltration (see section 4). We confirmed it with the author of the paper. There is however a typo at the last paragraph of section 4: 

"The calculated value for Idesign on a per square foot basis for each benchmark building is then 0.2016 cfm/ft2 for all above ground envelope area", it should be "The calculated value for Idesign on a per square foot basis for each benchmark building is then 0.2016 cfm/ft2 for all above ground wall area".

I did not find the 0.2016 cfm/ft² value in the Section 3.2.5 of the [ASHRAE 90.1-2010 PRM Reference Manual](https://www.pnnl.gov/main/publications/external/technical_reports/PNNL-25130.pdf) provides a different assumption.  0.4 cfm/ft2 of exterior building enclosure area at testing conditions (75 Pa) is aligned with the unit of 1.8 cfm/ft2 per above-grade envelope surface area. 

The idf snippet that you pointed out show an infiltration rate of 0.001024 m3/s/m2 per exterior area because that particular zone does not have a roof, there is a plenum above it. If you look at the apartment prototypes, you'll notice the difference that I mentioned.

mdahlhausen:
@asparke2 I reverted the code back to the original method per the comments above.  This is good to merge.

Issue #560, opened 2018-12-17
Create DOE Prototype Building BCL measure adds duplicate InternalMass objects


mldichter:
The idf created for the Full Service Restaurant has InternalMass objects duplicated for the zones. The previous version of this idf before being updated to Title 24 in the BCL measure has only Dining Mass and Kitchen Mass objects with the same fields as here. This probably occurred because the code added a Dining Mass and Kitchen Mass object multiple times, in which case openstudio appends a number to the end of the object name.

Running the idf with and without the duplicate InternalMass objects removed used different amounts of energy. The duplicate objects should be removed from the outputted idf file for correct behavior.

InternalMass,
    Dining Mass,             !- Name
    InteriorFurnishings,     !- Construction Name
    Dining ZN,               !- Zone Name
    743.49614298;            !- Surface Area {m2}

InternalMass,
    Dining Mass 1,           !- Name
    InteriorFurnishings,     !- Construction Name
    Dining ZN,               !- Zone Name
    743.49614298;            !- Surface Area {m2}

InternalMass,
    Kitchen Mass,            !- Name
    InteriorFurnishings,     !- Construction Name
    Kitchen ZN,              !- Zone Name
    278.8104884;             !- Surface Area {m2}

InternalMass,
    Kitchen Mass 1,          !- Name
    InteriorFurnishings,     !- Construction Name
    Kitchen ZN,              !- Zone Name
    278.8104884;             !- Surface Area {m2}

InternalMass,
    Kitchen Mass 2,          !- Name
    InteriorFurnishings,     !- Construction Name
    Kitchen ZN,              !- Zone Name
    278.8104884;             !- Surface Area {m2}

InternalMass,
    Kitchen Mass 3,          !- Name
    InteriorFurnishings,     !- Construction Name
    Kitchen ZN,              !- Zone Name
    278.8104884;             !- Surface Area {m2}

mldichter:
Error also present in the Warehouse model, which makes a big difference in the simulation from the amount of mass stored there.

!-   ===========  ALL OBJECTS IN CLASS: INTERNALMASS ===========

InternalMass,
    Zone1 Office Mass,       !- Name
    InteriorFurnishings,     !- Construction Name
    Zone1 Office ZN,         !- Zone Name
    473.759293161193;        !- Surface Area {m2}

InternalMass,
    Zone2 Fine Storage Mass, !- Name
    InteriorFurnishings,     !- Construction Name
[...]

PR #577, opened 2019-01-29
Change conditioned spaces checking when adding internal mass

mdahlhausen:
Fixes issue #560

The model was adding internal mass objects to conditioned spaces per a single-use method model_find_conditioned_space_names.  That method gets all system_to_space_map json objects and returns a vector of all spaces under each system in the system_to_space_map.  Sometimes, multiple systems serve the same conditioned space (e.g. unit heater and unit ventilator), so the space would get multiple internal mass objects.

This commit removes the model_find_conditioned_space_names method and instead checks whether a space is conditioned with the space_heated? and space_cooled? methods.


asparke2:
@mdahlhausen I think it could be because the large office and large office detailed tests had run_models set to true accidentally during the recent detailed offices merge.  I've made a branch setting that to false to confirm/refute this hunch.

Issue #503, opened 2018-09-27
Create DOE Prototype Building BCL measure has ceilings with exposed insulation and no drywall


mldichter:
Input:
LargeHotel
90.1-2013
ASHRAE 169-2006-2A
Not Applicable

The exposed building material in the ceilings is raw insulation, the drywall is missing from the construction.

Construction,
    LargeHotel Interior Ceiling,  !- Name
    CP02 CARPET PAD,         !- Outside Layer
    100mm Normalweight concrete floor,  !- Layer 2
    Nonres_Floor_Insulation; !- Layer 3

mdahlhausen:
Fixed with #748

Issue #500, opened 2018-09-24
Create DOE Prototype Building BCL measure has carpet in most rooms, kitchens, bathrooms, gyms, etc


mldichter:
Input:
SecondarySchool
90.1-2013
ASHRAE 169-2006-2A
Not Applicable

BuildingSurface:Detailed,
  Gym_ZN_1_FLR_1_Floor,                   !- Name
  Floor,                                  !- Surface Type
  Typical Insulated **_Carpeted_** 8in Slab Floor, !- Construction Name
  Gym_ZN_1_FLR_1 ZN,                      !- Zone Name
  Ground,                                 !- Outside Boundary Condition
  ,                                       !- Outside Boundary Condition Object
  NoSun,                                  !- Sun Exposure
  NoWind,                                 !- Wind Exposure
  ,                                       !- View Factor to Ground
  ,                                       !- Number of Vertices
  38, 52, 0,                              !- X,Y,Z Vertex 1 {m}
  38, 0, 0,                               !- X,Y,Z Vertex 2 {m}
  0, 0, 0,                                !- X,Y,Z Vertex 3 {m}
  0, 52, 0;                               !- X,Y,Z Vertex 4 {m}

BuildingSurface:Detailed,
  Kitchen_ZN_1_FLR_1_Floor,               !- Name
  Floor,                                  !- Surface Type
  Typical Insulated **_Carpeted_** 8in Slab Floor, !- Construction Name
  Kitchen_ZN_1_FLR_1 ZN,                  !- Zone Name
  Ground,                                 !- Outside Boundary Condition
  ,                                       !- Outside Boundary Condition Object
  NoSun,                                  !- Sun Exposure
  NoWind,                                 !- Wind Exposure
  ,                                       !- View Factor to Ground
  ,                                       !- Number of Vertices
  24, 9, 0,                               !- X,Y,Z Vertex 1 {m}
  24, 0, 0,                               !- X,Y,Z Vertex 2 {m}
  0, 0, 0,                                !- X,Y,Z Vertex 3 {m}
  0, 9, 0;                                !- X,Y,Z Vertex 4 {m}

BuildingSurface:Detailed,
  Bathrooms_ZN_1_FLR_1_Floor,             !- Name
  Floor,                                  !- Surface Type
  Typical Insulated **_Carpeted_** 8in Slab Floor, !- Construction Name
  Bathrooms_ZN_1_FLR_1 ZN,                !- Zone Name
[...]

PR #579, opened 2019-01-29
update construction properties

mdahlhausen:
- add gypsum board to LargeHotel InteriorFloor / InteriorCeiling constructions
- add additional GroundContactFloor constructions to allow for uncarpeted slab constructions


mdahlhausen:
Fixes issue #503 
Adds additional constructions to be used for future fix of issue #500

mdahlhausen:
A reasonable change but not necessary.  Include once added to the E+ base models from PNNL.

PR #584, opened 2019-02-01
Retrieves the heating section type for PSZ-AC

lymereJ:
The proposed changes include an instance variable that is used to retrieve the heating section type before looking up the cooling coil efficiency.

### DIFF
--- a/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
+++ b/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
@@ -2278,9 +2278,13 @@ class Standard
         return []
       end
 
+      # store the heating section type, used for looking up the correct efficiency for unitary equipment
+      @htg_coil_type = nil
+
       # create heating coil
       case heating_type
       when 'NaturalGas', 'Gas'
+        @htg_coil_type = 'Gas'
         htg_coil = create_coil_heating_gas(model,
                                            name: "#{air_loop.name} Gas Htg Coil")
       when 'Water'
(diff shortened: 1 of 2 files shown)


mdahlhausen:
@lymereJ I'm trying to understand why this change is needed.  I don't see how it changes the outcome of the `coil_dx_heating_type(coil_dx)` function.

PSZ-AC systems are created as air loops, so this section of code in `Standards.CoilDX.rb` applies:
```
# If on an AirLoop
if coil_dx.airLoopHVAC.is_initialized
  air_loop = coil_dx.airLoopHVAC.get
  htg_type = if !air_loop.supplyComponents('OS:Coil:Heating:Electric'.to_IddObjectType).empty?
               'Electric Resistance or None'
             elsif !air_loop.supplyComponents('OS:Coil:Heating:Gas'.to_IddObjectType).empty?
               'All Other'
             elsif !air_loop.supplyComponents('OS:Coil:Heating:Water'.to_IddObjectType).empty?
               'All Other'
             elsif !air_loop.supplyComponents('OS:Coil:Heating:DX:SingleSpeed'.to_IddObjectType).empty?
               'All Other'
             elsif !air_loop.supplyComponents('OS:Coil:Heating:Gas:MultiStage'.to_IddObjectType).empty?
               'All Other'
             elsif !air_loop.supplyComponents('OS:Coil:Heating:Desuperheater'.to_IddObjectType).empty?
               'All Other'
             elsif !air_loop.supplyComponents('OS:Coil:Heating:WaterToAirHeatPump:EquationFit'.to_IddObjectType).empty?
               'All Other'
             else
               'Electric Resistance or None'
             end
end
```

For a PSZ-AC system, this will return `'All Other'` except for when the air loop contains a `CoilHeatingElectric`, in which case it will return `'Electric Resistance or None'`.  The coils are [created here](https://github.com/NREL/openstudio-standards/blob/Fix/heating-section-type-for-eff-lookup/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb#L2281-L2323).  A `CoilHeatingElectric` is created if `heating_type` is `'Electric'`,`'Electricity'`, or `nil`.

Looking at your code, I'm not sure it changes the output of the function.

If a dummy coil is added, it triggers `'Electric Resistance or None'` but doesn't get changed to `'All Other'` because `@htg_type` is initialized to nil and doesn't trigger this block:
```
# For PSZ-AC only, retrieve the heating section type
unless @htg_coil_type.nil?
  if @htg_coil_type != 'Electric'
    htg_type = 'All Other'
  end
end
```

Is there something I'm missing?  Otherwise I'll just close this request.

lymereJ:
@mdahlhausen, the reason behind the proposed changes is that as it is, for PSZ-AC system with gas heat, `htg_type` is always set to `Electric Resistance or None`. [This block of code](https://github.com/NREL/openstudio-standards/blob/Fix/heating-section-type-for-eff-lookup/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb#L2326-L2339) is the reason why this is happening: PSZ-AC systems don't typically have supplemental heating coils so an electric heating coil with no capacity is created. The block of code you pointed out will see the electric heating coil and assign the `htg_type` variable to `Electric Resistance or None`.

If you don't like the changes that I'm proposing, we could change the order of the if/elsif statements so that the code looks first for a gas/fuel heating coil and if doesn't find any then looks for an electric coil. Alternatively we could check if the electric coil capacity is not equal to 0.

Let me know what you think, I can quickly revise the code and recommit. Thanks!

mdahlhausen:
@lymereJ changing the order looks like the best approach

mdahlhausen:
@lymereJ I updated the branch so it checks for the electric coil last.  Does this solve the issue?

lymereJ:
It should, I can check in a little bit.

lymereJ:
@mdahlhausen, I just tried the modification, it solves the issue.

PR #595, opened 2019-02-19
SWH DOE Prototypes Update

lymereJ:
List of the proposed changes:
* Updated water heater off cycle loss coefficient to ambient temperature
* Updated Ambient Temperature Schedule/Indicator/Thermal Zone
* SWH unit test


lymereJ:
I will revise the `water_heater_mixed_apply_efficiency` method instead of using hard-coded values for the coefficients.

PR #607, opened 2019-03-07
Feature/elevators

mpraprost:
- added new sheet in openstudio-standards for determining numbzer of elevators based on building type; changed code to pull elevator logic from this new sheet
- changed code such that elevator schedules are based off change in occupancy rather than pulled from openstudio-standards
- added elevator test for DEER models


mdahlhausen:
@mpraprost it looks like this pull request had la100 merged into it.  Since we don't want to merge la100 into master, I'm going to merge this into la100 instead and then cherry-pick commits to get the elevator stuff into master.

PR #605, opened 2019-03-06
Weather std update

mdahlhausen:
Adds in both ASHRAE 169-2006 and ASHRAE 169-2013 to the spreadsheet and the code base.
 - Includes updated city locations for each climate zone
 - Includes updated ground temperature for each climate zone
 - Updates the prototype models and tests to use ASHRAAE 169-2013.

Alternative solution to accommodate pull #572

PR #610, opened 2019-03-07
Regression and Simulation Tests for the DOE/90.1 Prototypes

lymereJ:
The proposed changes add regression and simulation tests for the DOE/901 prototypes. This first commit includes the structure of the tests. The result CSV file used for comparison and the *true* models for the regression test still need to be produced and added to the pull request.


lymereJ:
@asparke2, @mdahlhausen - the last build for this PR failed because of `test/doe_prototype/test_add_hvac_systems.rb`. I ran this test on my machine using my local branch which is the same as what is in the latest commit and it didn't fail. Am I missing something?

mdahlhausen:
@lymereJ I think there were some HVAC systems that started showing unmet hours with the weather location changes from the ASHRAE 169 merge.  Can you update / pull changes to those tests from master if there are any into this branch?

Otherwise, I'll pull down this branch and debug the HVAC system tests.

lymereJ:
@mdahlhausen I did an update from master on March 7. It looks like you merged #605 on the 6th, so it should be good, shouldn't it?

lymereJ:
@asparke2 - It looks like #663 did the trick!

PR #620, opened 2019-03-12
model_add_elevators refactor and tests

mdahlhausen:
From @mpraprost:
- added new sheet in openstudio-standards for determining numbzer of elevators based on building type; changed code to pull elevator logic from this new sheet
- changed code such that elevator schedules are based off change in occupancy rather than pulled from openstudio-standards
- added elevator test for DEER models

update of pull #607 to merge with master

once tests are complete, add the Elevators sheet to the master google spreadsheet.


mdahlhausen:
@asparke2 it seems like this is failing because of a regression test time?

asparke2:
@mdahlhausen it actually wasn't a timeout issue, there were tests with errors (not failures) that circleci didn't catch for whatever reason.  Looks like some necessary methods in Standard.ThermalZone.rb weren't included in commits originally.

PR #676, opened 2019-06-28
change design spec oa method to retain values for DCV

mdahlhausen:
If a zone or system did not require DCV, the values in the design specification outdoor air object were all converted to flow per area.  This means that DCV cannot be applied to the model in a future measure, as that information is lost.  This refactor changes the conversion to a flow per zone and sets the oa method to Flow/Zone to retain per person values for future DCV application.


mdahlhausen:
---
_Matt:_
Working on the IDSM DCV measure which may apply to ComStock as well.

Should this be ‘unless’ instead of ‘if’?:
```
if thermal_zone_demand_control_ventilation_required?(zone, climate_zone)
  thermal_zone_convert_oa_req_to_per_area(zone)
end
```
Measure scope: 
 - Should we restrict the kinds of HVAC systems that apply, e.g. only VAV? Or include DOAS systems as well?  Other system types?
 - If all zones have ventilation specified per area, should we convert it to per person?  Or require the model have per person values already specified?  The air_loop_hvac_apply_standard_controls method that we use in the protoype and create_typical_building methods/measures converts ventilation to per area (see code above).  Should we retain per person information as an additional property when the model is built in create typical building?  Otherwise we won’t capture per area ventilation minimums.
---
_Andrew:_
1.	Yes, I think this should be “unless”
a.	Add a new method to replace `thermal_zone_convert_oa_req_to_per_area` but the new method converts all numbers to a flow per zone and converts the Dsn Spec OA to “_Flow/Zone_” instead of “_Sum_”
b.	Need to change `thermal_zone_outdoor_airflow_rate `calculation to account for other Dsn Spec OA methods, especially “_Flow/Zone_”
2.	Since all the DEER ventilation rates are only cfm/area
a.	Convert flow/area values to flow/area and flow/person using 0.15 cfm/ft2 assumption from T24
b.	Then apply DCV

mdahlhausen:
The space type ventilation values get set to empty during OpenStudio translation if an outdoor air method that doesn't use them is selected.  We'll need to find another way to retain ventilation rates.

mdahlhausen:
closing this for now.  We still need to do this for some measures that rely on DCV, but assuming 20 cfm/person in the measure is a lot easier than refactoring openstudio-standards.

mdahlhausen:
This issue is addressed with #919 which keeps the original values for the design specification outdoor air.  Deleting this branch because the changes are no longer necessary.

PR #677, opened 2019-07-11
Update OpenStudioStandards::get_run_env to match OpenStudio::Extensio…

macumber:
…n::Runner::get_clean_env


macumber:
@asparke2 we should discuss this to make sure it will work for you

macumber:
Thanks @asparke2 !

PR #675, opened 2019-06-27
Daylighting controls input #3

lymereJ:
This is a follow up on #654 and #655 and addresses daylighting controls input in additional prototypes (Warehouse, Large Office, Medium Office, Small Office, Restaurants,and Standalone Retail). The tests should fail. Updated regression files and simulation results will follow.

PR #679, opened 2019-08-05
Enable doe regression tests

mdahlhausen:
Enables the DOE prototype model regression tests and improves logging to diagnose problems.

A major change since the inclusion of regression tests is updating weather data in alignment with ASHRAE 169-2013 (from ASHRAE 169-2006).  This causes some minor changes (+-3 tons on ~400 ton chillers).  [Details here](https://github.com/NREL/openstudio-standards/commit/95e5605413c503be1471abf7f54e1c673e0d091f).

Andrew: The legacy_idf_results.json is still using ASHRAE 169-2013.  We should add the ASHRAE 169-2013 results to it.  There are minor (+-2% changes in specific end uses, most <1%, and a few pump values <8%, largely all chiller-based systems).


asparke2:
@lymereJ, @mdahlhausen discovered that the DOE regression tests weren't actually enabled in the commit where you added them: https://github.com/NREL/openstudio-standards/pull/610.  This means that they haven't been running.  In order to get them running, we need to make an updated legacy results file that uses the updated 2013 weather file locations.  I think this is something that you can do?

lymereJ:
@asparke2, @mdahlhausen - Thank you for catching that! I did notice that the result comparison flag for the performance test was turned off when working on #675. I turned it back on then and noticed that the test failed. After looking into it, the reason is that we're creating the legacy results on a Linux based machine and CircleCI runs on Windows based machines. We noticed that while the input files are identical, several SizingPeriod simulations show different results. In these instances, the ERV included in the model is bypassed when it shouldn't be. We reported back to Mike Witte who identified the issue and posted it on [EnergyPlus' GitHub page](https://github.com/NREL/EnergyPlus/issues/7379). One way around it would be to use a `LockoutWithHeating` strategy instead of `NoLockout`. Our current prototype models don't use `NoLockout` but the ones created by the gem do. We are planning to update this soon and it should hopefully resolve the SizingPeriod run discrepancies.

While reviewing the proposed code changes I noticed that you turned off some code versions from the performance test, is that to speed up the build? I also noticed that you changed the value of the `acceptable_error_percentage` variable, could you please let us know why?

Doesn't the `test_regression.rb` need to be added to the `circleci_test.txt` file? I failed to add it in #610 but it looks like it's also not part of the proposed changes.

mdahlhausen:
@lymereJ we were working on the `test_performance.rb ` file.  Since we enabled an object-by-object comparison to .osm files, we are restricting to .osm files available in the 'regression_models' directory.

Should `test_performance.rb` be only for the 4A / 90.1-2004, 90.1-2013 models available in `regression_models` and `test_regression.rb` test everything?  We still need to update the `legacy_idf_results.json` file to include ASHRAE 169-2013 results.

Are you available for a call today to work through what should be in the regression test updates?

PR #687, opened 2019-08-29
Radiant systems

mdahlhausen:


lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -197,6 +197,8 @@ def model_add_hw_loop(model,
   # @param num_chillers [Integer] the number of chillers
   # @param condenser_water_loop [OpenStudio::Model::PlantLoop] optional condenser water loop for water-cooled chillers.
   #   If this is not passed in, the chillers will be air cooled.
+  # @param add_waterside_economizer [Bool] If true, will add an ideal heat exchanger between the condenser loop
+  #   and plant loop to enable waterside economizer when wet bulb temperatures allow
```
asparke2: Modify documentation to say whether this is an integrated or non-integrated.  I assume you tested that the controls/modeling piece works as expected.  If not, check the code against the measure here (not on BCL) because I tested the modeling results in this measure quite a bit: https://github.com/NREL/OpenStudio-measures/blob/develop/eto_dev/waterside_economizer/measure.rb.  Do you think it's worth making a separate 'plant_loop_add_waterside_economizer` method with options for integrated or non-integrated and calling that from this spot?

mdahlhausen: I've add integrated and non-integrated options based on the ptool measures.
I decided to keep it as model_add_waterside_economizer in the Prototype.hvac_systems file instead of plant_loop_add_waterside_economizer in the plant methods.

mdahlhausen: Add some additional setpoint managers and set minimum temperature diff for HX to 0F for proper operation.  Seems like this issue is still occurring:
https://github.com/NREL/EnergyPlus/issues/5626

mdahlhausen:
@asparke2 if the tests pass this is good to merge

asparke2:
@mdahlhausen I don't want to encourage the practice of adding more code to already-merged branches, so if/when @carlosduarteroa starts to add more control and sizing schemes he'll have to create a new branch off of master.

PR #686, opened 2019-08-29
align ashrae 90.1 prototypes simulation control settings

leijerry888:
The following objects are aligned between osstd prototypes and PNNL maintained doe prototypes:
* timestep
* sizing:parameters
* convergencelimits
Changes are mostly made by modifying geometry osm files.

PR #694, opened 2019-09-11
fix HighRiseApartment to use same heat pump loop for all WSHPs

mdahlhausen:
@lymereJ
The openstudio prototype model for HighRiseApartment is creating a heat pump loop for every WSHP unit rather than having one heat pump loop serve all WSHPs as in the DOE Prototype .idf files.

This is fixed easily by changing [this line in Prototype.Model.hvac.rb](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/common/objects/Prototype.Model.hvac.rb#L119) from 
`heat_pump_loop = model_add_hp_loop(model)`
to
`heat_pump_loop = model_get_or_add_heat_pump_loop(model)`

If you are ok with this change, either you or I can update the regression models and expected performance results and merge it.


mdahlhausen:
@leijerry888 @lymereJ yes, we should merge #686 first, then update this branch from master, otherwise there will be a merge conflict in the regression models and performance results.

mdahlhausen:
@leijerry888 @lymereJ I merged in #686 so we are ready to update the regression models and performance results now.

PR #701, opened 2019-09-18
add WSHP data and methods

mdahlhausen:
- Adds water_source_heat_pumps and water_source_heat_pumps_heating sheets to openstudio-standards data
- Adds methods to apply WSHP coil efficiencies based on standards data


mdahlhausen:
@lymereJ and @leijerry888 this adds WSHP efficiency methods and data.  It changes the highrise apartment and large office models.  Please review the changes and then either you or I can regenerate the regression models and performance results.

jiangithub:
@mdahlhausen 

> @lymereJ and @leijerry888 this adds WSHP efficiency methods and data. It changes the highrise apartment and large office models. Please review the changes and then either you or I can regenerate the regression models and performance results.

We will review this. For Large Office, we used Water Source Heat Pump to emulate the water-cooled DX units. The efficiency inputs for Computer Room Air Conditioner (CRAC) Units should follow Table 6.8.1-11 Floor Mounted Air Conditioners and Condensing Units Serving Computer Rooms—Minimum Efficiency Requirements (90.1-2016 example). The requirements are differences between 90.1-2010/2013 and 2016. There was no applicable requirement in 90.1-2004/2007 and we should use ASHRAE Standard 127. We can discuss this at our next call.

mdahlhausen:
@jiangithub 

My thinking is there are two issues here:
1) applying WSHPs 90.1 efficiency values
2) how to model CRAH units in the large office data center.

For 1) This will better align the HighriseApartment prototype and won't make the LargeOffice model worse - assigning 90.1 efficiency values to the WSHP equipment is better than the default.

For 2) There is a [model_add_crah]( https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb#L3091-L3243) method in Prototype.hvac_systems to apply computer room air conditioners.  We could use this method for the LargeOffice data center, or keep the WSHP model and apply efficiency/equipment changes in the Prototype.LargeOffice [model_custom_hvac_tweaks](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/common/buildings/Prototype.LargeOffice.rb#L6-L42) method.

Currently, other data changes are held up because of the addition to the _water_source_heat_pumps_ and _water_source_heat_pumps_heating_ sheets to the OpenStudio_Standards-ashrae_90_1 spreadsheet.  They aren't listed in the [unique_properties](https://github.com/NREL/openstudio-standards/blob/master/data/standards/manage_OpenStudio_Standards.rb#L83) method of manage_OpenStudio_Standards.rb on master, so the json export fails.

Given that, I suggest we regenerate the regression models and merge this, and then figure out how to better align the LargeOffice data center CRAC units as a separate issue and pull request.

jiangithub:
@mdahlhausen 
For 1) @leijerry888 will review the WSHP curves and the COP. 
For 2) model_add_crah method is not an intended HVAC system for the data center in the large office prototype. We may use model_custom_hvac_tweaks method but we'll have to develop new tweaks. So, I agree that we need to create a new issue and pull request for 2) and don't have to resolve it in this PR. 
After 1) is verified, @leijerry888 will generate the regression results and models and add them to this branch.

lib/openstudio-standards/standards/Standards.CoilCoolingWaterToAirHeatPumpEquationFit.rb
```diff
@@ -0,0 +1,109 @@
...
+      return successfully_set_all_properties
+    end
+
+    # Get the minimum efficiency standards
+    cop = nil
+
+    # If specified as EER (heat pump)
+    unless coil_props['minimum_full_load_efficiency'].nil?
+      min_eer = coil_props['minimum_full_load_efficiency']
+      cop = min_eer / 3.4121416
```
leijerry888: @mdahlhausen , In the generated idf, I did a spot check, COP for Coil:Cooling:Watertoheatpump:Equationfit seems too low. It looks like this may be the cause. Please see the following I quoted from @jiangithub 

"For packaged (with fan and coils) heating and cooling HVAC equipment, the industry standards for efficiency rating includes the fan energy inputs in their efficiency rating, such as EER, SEER, heating COP. When we model such HVAC systems in EnergyPlus, we have separate fan and coil objects, in which their efficiency inputs are separate. So, the modeled COP input for a DX cooling coil is for the compressor only and does not include fan. Therefore, we should convert the efficiency metrics based on the industry standards to the modeled efficiency inputs. 

PNNL Method- From Thornton et al 2011. 

Thornton B.A., M.I. Rosenberg, E.E. Richman, W. Wang, Y. Xie, J. Zhang, and H. Cho, et al. 2011. Achieving the 30% Goal: Energy and Cost Savings Analysis of ASHRAE Standard 90.1-2010. PNNL-20405. Richland, WA: Pacific Northwest National Laboratory.

COP= (EER/3.413 + R) / (1-R)

EER = -0.0182 * SEER2 + 1.1088 * SEER"

leijerry888:
I did ONE quick spot check. The curve numbers and heating coil COP looks good. However, the cooling coil COP does not agree with PNNL models. I created a review around the suspicious code with more details of our modeling strategy.

@mdahlhausen @jiangithub @lymereJ , if it is preferred to generate the regression models and performance test expected results before figuring out this issue, let me know and I will commit them.

KaiyuSun:
@mdahlhausen @jiangithub @leijerry888 I added the functions model_add_crac and model_add_crah for the data center prototype models. model_add_crac is adding **computer room air conditioner**, served by **air cooled** DX cooling coil. model_add_crah is adding CRAH system, served by chilled water system. Since the data centers in the large office model use **WSHP** system, which uses different components from the above two systems, I think neither of these two functions are the right solution for it. Let me know if you have further questions.

lib/openstudio-standards/standards/Standards.CoilCoolingWaterToAirHeatPumpEquationFit.rb
mdahlhausen: @leijerry888 I [changed the cooling coil](https://github.com/NREL/openstudio-standards/commit/338823e89ba62f6af2d6fa9d65394b173d9091f2) to use the [eer_to_cop](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/common/objects/Prototype.utilities.rb#L289-L311) method that removes the fan from the cop accounting.  Should we do this for the heating coil as well?  I was hesitant to apply it there because that method [cop_heating_to_cop_heating_no_fan](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/common/objects/Prototype.utilities.rb#L246-L253) assumes a COPH rated at 47F, rather than 68F for water-to-air coils.  Do you know what efficiency adjustment method we should use?

leijerry888: Thanks @mdahlhausen . The heating coil COP looks good probably because 4.2 is the only option in there. @jiangithub , could you please confirm the answer to @mdahlhausen 's question? In the meantime, I will pull the new commit and test it.

jiangithub: @mdahlhausen We didn't convert the rated heating COP in 90.1 to a COP input in the EnergyPlus Highrise Apartment. We took some manufacture specification data and developed a set of curves to make the COP and curve inputs to match the manufacture data and meet the COP in 90.1. 
I am not sure if [cop_heating_to_cop_heating_no_fan](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/common/objects/Prototype.utilities.rb#L246-L253) method is an alternative way to do so. For the purpose of aligning the OpenStudio Standards prototypes with the EnergyPlus ones, let's use the same curves and use the COP in 90.1 and do not make any conversion.

jiangithub:
> @mdahlhausen @jiangithub @leijerry888 I added the functions model_add_crac and model_add_crah for the data center prototype models. model_add_crac is adding **computer room air conditioner**, served by **air cooled** DX cooling coil. model_add_crah is adding CRAH system, served by chilled water system. Since the data centers in the large office model use **WSHP** system, which uses different components from the above two systems, I think neither of these two functions are the right solution for it. Let me know if you have further questions.

@KaiyuSun Thank you for the clarification.

lib/openstudio-standards/standards/Standards.CoilCoolingWaterToAirHeatPumpEquationFit.rb
leijerry888: Results of a quick check: Comparing 2004 PNNL El_Paso v.s. 2004 OSSTD 3B: both have 80 coil:cooling:WaterToAirHeatPump:EquationFit objects. In PNNL model, 75/80 objects have COP of 3.86 and 5/80 have COP of 4.13; In OSSTD model, 59/80 objects have COP of 3.86 and 21/80 have COP of 4.13. I haven't looked into the cause.

leijerry888: @mdahlhausen , I just discussed with @jiangithub , we think the COP differences mentioned in my previous comment are expected due to expected capacity differences between the PNNL and OSSTD models. As a sanity check at this moment, we think it is good as long as the calculated COP falls in one of the possible choices calculated based on the capacity.  I will check more model years and climate zones, and if all are good, I will commit the performance test results and regression models.

mdahlhausen: @leijerry888 thanks for looking into this.  I expect the differences are from capacity.

For the curves - I didn't add coefficient methods or data yet, so these models are using the default OpenStudio/EnergyPlus coefficients.  Openstudio handles the curve coefficients for the WaterToAirHeatPump:EquationFit objects a bit differently by having coefficient fields in the object itself rather than as linked curves.  I can add the data/methods if you want that done in this pull request.

leijerry888: @mdahlhausen The coefficients used in OSSTD models seem to be not default values, but from [here](https://github.com/NREL/openstudio-standards/blob/30818329a8a2db5e2da374cd93d4865c0e8fe7f1/lib/openstudio-standards/prototypes/common/objects/Prototype.CoilCoolingWaterToAirHeatPumpEquationFit.rb#L44-L59) and [here](https://github.com/NREL/openstudio-standards/blob/30818329a8a2db5e2da374cd93d4865c0e8fe7f1/lib/openstudio-standards/prototypes/common/objects/Prototype.CoilHeatingWaterToAirHeatPumpEquationFit.rb#L44-L53). These values match values in PNNL models. I am guessing it's because when creating these objects at [here](https://github.com/NREL/openstudio-standards/blob/4b4ae4b60ae9fc59bcd5b7a334c080ede24bfc15/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb#L2836-L2841), the default type (i.e. nil) is passed in instead of "OS default"

mdahlhausen: great!

leijerry888: Checked all COP instances of OSSTD high rise apartment models. Cooling coils set COP to be one of [4.46473377194151, 3.86542098393842, 4.1317822230509, 4.19837253282902], all heating coils use COP of 4.2

These value match PNNL WSHP coil efficiency table values. Issue resolved. Regression models and performance test commit will come shortly

mdahlhausen:
@asparke2 @leijerry888 the outpatient error seems to be some nondeterministic failure on the CircleCi testing machine.  Andrew and I don't get it when we run the test locally on a linux container.  We're good to merge this since that is the only failure.

leijerry888:
@mdahlhausen I re-run the performance test models and the outpatient error is gone. Not sure why at this moment, will investigate later as we also encountered some earlier instances of running same model but getting different results for performance tests.

PR #702, opened 2019-09-24
Fix front entry heater control EMS in the standalone retail prototype

lymereJ:
The proposed changes fix the front entry heater control. Currently, the EMS turn off the heater if the OAT is greater than 45 deg. F but it doesn't turn it back on when the OAT falls back below 45 deg. F.

Once #701 is merged, this branch will have to be updated from master. We'll regenerate the performance test results and the regression models then.


lymereJ:
Good to be merged even though one test has failed.

PR #705, opened 2019-09-30
correct DOAS method to enable DCV

mdahlhausen:
This fix does two things:
1) Sets the zone DOAS air terminal as the 1st priority for zone heating and cooling so that it's effect on the zone air is included
2) Sets the sequential cooling fraction and sequential heating fraction to zero so that the DOAS air terminal is only expected to provide ventilation, not zone conditioning.  In the case of economizing, the sequential cooling fraction is set to 1.0 to enable economizing, and is mutually exclusive of DCV.


mdahlhausen:
@asparke2 wait to merge this until after @pflaumingo reviews it

mdahlhausen:
note: the tests fail because the .setSequentialCoolingFraction method was added in OpenStudio v2.8.0, and the testing framework is using v2.7.1.

mdahlhausen:
@asparke2 I made a silly error - forgot to remove the common from the last item in the testing array when I commented out the last item.  should be good to remerge.

PR #710, opened 2019-10-09
Round outpatient min airflow fraction

leijerry888:
We observed an issue of generating and running performance tests (sizing) models with the same setting gives different results. This has been observed both from the process of generating expected performance test results within PNNL and from running performance tests on CircleCI. By checking idf model differences of runs for these different results, we pinpoint the issue as a lack of rounding for constant minimum airflow fraction of single duct reheat vav in some scenarios. A rounding is added accordingly.


leijerry888:
rebuilt circle ci tests multiple times, one succeeded, others failed for various reasons: "Fractional 1" object, time out, dependency installation errors. Should be good to merge

PR #714, opened 2019-10-16
Different results between 7A/7B, and 8A/8B

leijerry888:
We found that OSSTD offers different models and thus different simulation results for climate zone 7A vs 7B, and 8A vs 8B.

The issue is identified to be incorrect climate zone values at https://github.com/NREL/openstudio-standards/blob/935ad6448b558ab938ab291f74157722dbdacd57/lib/openstudio-standards/standards/ashrae_90_1/ashrae_90_1_2010/data/ashrae_90_1_2010.ground_temperatures.json#L550
(not only for 2010, but also other standard years, and for all prototypes besides LOC550)

This causes a mismatch of 7A and 8A when create group temperature objects and leads the creation of such objects with default temperature values at https://github.com/NREL/openstudio-standards/blob/d555eda54d026c598727fd758335c974b7bd73f7/lib/openstudio-standards/weather/Weather.Model.rb#L195-L207

Related json files are modified.
The following cells in Google spreadsheet is also updated.
https://docs.google.com/spreadsheets/d/1MYQGPd3WFaaSHGnwcmYoNi7TLDgE76LjecTEkY0vGco/edit#gid=2941906&range=C1612:C1683
https://docs.google.com/spreadsheets/d/1MYQGPd3WFaaSHGnwcmYoNi7TLDgE76LjecTEkY0vGco/edit#gid=2941906&range=C1788:C1859


leijerry888:
Performance test results updated. No need to update regression test models.
Minor differences in performance tests csv all over the place. I assume they are caused by changes made by #711 . 
All differences larger than 0.001% seem to be in 7A and 8A as expected.

leijerry888:
@mdahlhausen I was talking about the minor differences for all models when `git diff da68213 a351eb1 data/standards/test_performance_expected_dd_results.csv`

Is the performance test code checking all climate zones? Because if it does, then I would expect performance test failures in da68213 for 7A and 8A models.

leijerry888:
@mdahlhausen 

I think this answers my last question:
https://github.com/NREL/openstudio-standards/blob/90361bbd21652d6b27545142d20735ee018c4aae/test/doe_prototype/test_performance.rb#L22-L41

leijerry888:
I took out the "In Progress" tag. As long as the new performance test results look good, this should be ready to merge.

Issue #688, opened 2019-09-04
Issue with Internal Variables and Automated Baseline Generation


lymereJ:
When using the `model_create_prm_baseline_building` method on a model with `EnergyManagementSystem:InternalVariable` objects, the sizing run failed because objects related to the `EnergyManagementSystem:InternalVariable` (in that case a `Controller:OutdoorAir` object) was removed from the model by the measure.

mdahlhausen:
To further specify this problem - the idea is that when generating the baseline HVAC system, the model doesn't remove EnergyManagementSystem objects related to the proposed HVAC.

This is subset of a larger problem of orphaned EnergyManagementSystem objects.  Perhaps it makes sense to delete all EnergyManagementSystem objects as part of the method?

mdahlhausen:
@lymereJ I'm trying to recreate this with the 'Create DOE Prototype Building' measure in combination with the 'Create Baseline Building' measure, but it seems that the single zone controls aren't being applied in the .osm files in many instances where they exist in the .idf files.  Can you recommend a template, building type, and climate zone combination that exhibits this error?

lymereJ:
I was also using these two measures, here's an example: small hotel, 90.1-2004, and 5B.

mdahlhausen:
Fixed with https://github.com/NREL/openstudio-standards/pull/716

PR #716, opened 2019-10-22
remove ems objects when calling model_create_prm_baseline_building

mdahlhausen:
If a user makes a prototype HVAC system, or otherwise uses EMS objects in their model, these may get orphaned when running the prm_baseline method which removes all HVAC and service water heating.

This method removes EMS objects so that the baseline model can be made cleanly.

Intends to fix [NREL/OpenStudio issue #3279](https://github.com/NREL/OpenStudio/issues/3279) and #688.


MatthewSteen:
Note that #3279 is about the baseline model having orphaned EMS objects, not the proposed.

PR #757, opened 2020-02-24
don't add economizers to unitary systems

### LINKED ISSUE #756: Don't apply economizers to unitary systems
Unitary HVAC systems generally don't have outdoor air, so they can't have an economizer.

The `air_loop_hvac_economizer_required?` method doesn't check if an air loop is a unitary system, and so residential HVAC system types in openstudio-standards get thrown this error:
`OpenStudio.logFree(OpenStudio::Error, 'openstudio.prototype.Model', "#{air_loop.name} is required to have an economizer, but it has no OA system.")`

The fix is to add a check in the `air_loop_hvac_economizer_required?` method to skip unitary system types.

mdahlhausen:
Fixes #756

### DIFF
--- a/lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb
+++ b/lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb
@@ -827,8 +827,8 @@ class Standard
   def air_loop_hvac_economizer_required?(air_loop_hvac, climate_zone)
     economizer_required = false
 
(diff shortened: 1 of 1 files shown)


jiangithub:
Single-zone systems, such as PSZ-AC and PSZ-HP are unitary systems. They serve commercial spaces and may have air economizers. We should not use "Unitary" as a keyword to disable economizers.

PR #758, opened 2020-02-24
Comstock fixes

mpraprost:
merging a series of comstock-related fixes into master branch


mdahlhausen:
@lymereJ @jiangithub This merge mostly impacts the comstock standards sets, but we did make a change regarding the `lighting fraction replaceable` data field to explicitly set it.

We are failing 2 regression/performance tests, both for the midrise apartment.
[See failures here](https://circleci.com/gh/NREL/openstudio-standards/3479#tests/containers/2)

On `master`, the apartment lights replaceable field is set to [`null`](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/standards/ashrae_90_1/ashrae_90_1_2004/data/ashrae_90_1_2004.spc_typ.json#L5862), while on the `comstock_fixes` branch it is to [1.0](https://github.com/NREL/openstudio-standards/blob/comstock_fixes/lib/openstudio-standards/standards/ashrae_90_1/ashrae_90_1_2004/data/ashrae_90_1_2004.space_types.json#L5791).  The `null` value in master is causing this field to be set to 0.0 in the regression and performance models.  
The PNNL DOE prototype .idf files have no input for this field, and the EnergyPlus IDD shows this defaults to 1.0.  The energy use is different, presumably because of interactions with the daylighting script (LPDs are identical). Are you ok with updating the performance and regression models to use the 1.0 value?

jiangithub:
@mdahlhausen Do you intend to enable daylighting controls for comstock apartment prototypes? Technically, it is very rare apartment buildings to use daylighting controls. The public spaces like corridor might be applicable but they usually don't have much access to windows.

mdahlhausen:
@jiangithub I agree that installed daylighting controls are rare in residences; the occupant is the daylighting control.  However, I think this should be handled by a separate method in openstudio-standards, rather than fixing it so that no daylighting controls will work via the fraction replaceable field in the EnergyPlus object.

mpraprost:
@mdahlhausen has it been decided whether we should change the lighting fraction replaceable back to null, or add something to standards in order to fix these failing tests?

mdahlhausen:
@mpraprost I'll ask about this at our check-in with PNNL tomorrow

mdahlhausen:
After copy over master onto this branch and then resolving all merge commits (see last 4 commits), there aren't commits on the comstock_fixes branch that change things.  Seems like the .json files already got rebuilt on master, and most other fixes were merged into comstock_fixes along the way instead of the other way around.  Therefore I'm closing this pull request.  Use master; it has everything this has and more.

PR #762, opened 2020-03-03
fixing schedule inconsistencies with constant/hourly schedules

mpraprost:
- changed schedule type from constant to hourly for schedules with hourly values
- got rid of extra schedule values in constant schedules
- added missing values in hourly schedules

fyi - some new college model schedules got wrapped into this commit when i rebuilt the jsons


mdahlhausen:
Fixes #762

mdahlhausen:
@mpraprost this is throwing an error in LargeOffice-90.1-2013.

mpraprost:
@mdahlhausen this looks like a result of me changing the schedule type from constant to hourly, because the schedule had hourly values. any ideas why this would break the test?

mdahlhausen:
@mpraprost I checked the .idf copy of the PNNL prototypes and the openstudio-standards test regression model was incorrect.  I fixed the regression model so this should pass now.

PR #771, opened 2020-03-06
add an exception for apartment spaces to space_daylighting_control_required

mdahlhausen:


mdahlhausen:
@lymereJ can review and if regenerate the regression and performance results if good?
I can update regression models but wasn't sure what method you use to re-create the performance results.  This only impacts MidriseApartment and HighriseApartment.

PR #778, opened 2020-03-10
Update large office prototype data-center/data-closet equipment schedule

lymereJ:
Update large office prototype data-center/data-closet equipment schedule.


lymereJ:
Good to know, I'll revise the changes. Thanks!

PR #791, opened 2020-03-13
replace sql lookup for construction visible transmittance

mdahlhausen:
Use the construction visible transmittance instead of an sql lookup.

@asparke2 or @lymereJ or @DavidGoldwasser do you know why the space_sidelighting_effective_aperature and space_skylight_effective_aperature methods were using visible transmittance values from the sql file instead of just getting it from the construction property?  It necessitates a whole sizing run before applying daylighting controls and is causing some issues with our non-prototype model projects.

Will this fix be ok?


asparke2:
@mdahlhausen If I remember correctly, you can only get the VLT from the model for Constructions that use SimpleGlazing materials for layers.  If you have a complex fenestration using the normal glass/gas layers, I think E+ computes the VLT, which is why the sizing run is req'd.

mdahlhausen:
@asparke2 ok, then I'll revert the last commit.

What if we do a check to see if it has a visibleTransmittance value, and then if it doesn't, use the sql method?  Will that work?

asparke2:
Yeah, that’s probably a good choice.


  *   Andrew

PR #797, opened 2020-03-25
Bump dependencies (OpenStudio-update-gems)

jmarrec:
the rake & bundler lines are needed for 3.x*. For the rest including using pessimistic version constraints, up to you.

*Not really needed in the sense that it's a development_dependency, so it'll work for us, but since you'll be switching to ruby 2.5.5 soon, might as well update as well.

### DIFF
--- a/openstudio-standards.gemspec
+++ b/openstudio-standards.gemspec
@@ -23,24 +23,24 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'minitest-parallel_fork'
   spec.add_development_dependency 'ruby-progressbar'
   spec.add_development_dependency 'parallel_tests'
-  spec.add_development_dependency 'nokogiri', '<= 1.6.8.1'
-  spec.add_development_dependency 'bundler', '~> 1.9'
-  spec.add_development_dependency 'rake', '~> 10.0'
+  spec.add_development_dependency 'nokogiri', '~> 1.8.2'
+  spec.add_development_dependency 'bundler', '~> 2.1'
+  spec.add_development_dependency 'rake', '~> 12.3.1'
   spec.add_development_dependency 'yard', '~> 0.9'
-  spec.add_development_dependency 'rubocop', '0.68.1'
-  spec.add_development_dependency 'rubocop-checkstyle_formatter', '~> 0.1.1'
-  spec.add_development_dependency 'minitest-ci', '<= 5.10.3'
-  spec.add_development_dependency 'rubyXL', '3.3.8' # install rubyXL gem to export excel files to json
-  spec.add_development_dependency 'activesupport', '4.2.5' # pairs with google-api-client, > 5.0.0 does not work
-  spec.add_development_dependency 'public_suffix', '3.0.3' # fixing version of google-api-client dependency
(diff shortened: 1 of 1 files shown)


openstudio-standards.gemspec
```diff
@@ -23,24 +23,24 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'minitest-parallel_fork'
   spec.add_development_dependency 'ruby-progressbar'
   spec.add_development_dependency 'parallel_tests'
-  spec.add_development_dependency 'nokogiri', '<= 1.6.8.1'
-  spec.add_development_dependency 'bundler', '~> 1.9'
-  spec.add_development_dependency 'rake', '~> 10.0'
+  spec.add_development_dependency 'nokogiri', '~> 1.8.2'
+  spec.add_development_dependency 'bundler', '~> 2.1'
+  spec.add_development_dependency 'rake', '~> 12.3.1'
```
jmarrec: Should be `~> 13.0` probably

mdahlhausen:
@jmarrec does bundler need to be 2.0+?  If so, I don't think it will be possible to make openstudio-standards compatible with both OS 2.7.1 and OS 3.0.  @asparke2

jmarrec:
Yes it has to be 2.x. The reason for that is that the process of embedding gems into the CLI uses bundle itself and it isn't happy when you mix both 1.x and 2.x

But that doesn't mean it can't be compatible with os 2.x.
For most openstudio-related gems and the NREL/openstudio-gems project (and OpenStudio itself) we're generally providing a 2.9.X-LTS branch which should get only critical bug fixes.

I guess I'd need to understand what your use case is to advise here for backward compatibility (a lts branch, having conditional gemspec, or just doing nothing as requiring the local checked out source doesn't call the gemspec).

We should talk with @tijcolem too, here or on slack.

asparke2:
@jmarrec The reason we need to keep openstudio-standards compatible with _development_ using OS 2.7.1 (E+ 9.0) is because PNNL is still working on editing the DOE prototypes in openstudio-standards to match the official IDF versions, which are in E+ 9.0.  PNNL doesn't currently have the time/budget to update their official IDFs as quickly as OS & E+ move, and they need to be able to compare IDF results without the complication of accounting for differences E+ engine changes.

In order to preserve this compatibility, we need to make sure that the _development_ dependencies of openstudio-standards are compatible with Ruby 2.2.1.  This is why there were so many specific versions specified in the .gemspec.  For example, [bundler >= 2.0.0-pre1 requires Ruby >= 2.3.0](https://rubygems.org/gems/bundler/versions/2.0.0.pre.1).

I don't believe that it should matter which development dependencies openstudio-standards specifies, because when you embed the gems into the CLI, I believe you use the `--without test, development` option?

jmarrec:
@asparke2 https://github.com/jmarrec/openstudio-gems/blob/5fa1b33e651c835070cff39fffa2af1bd855771b/build_openstudio_gems.rb#L86

Not sure whether excluding `development` was considered or not.

asparke2:
@jmarrec In my opinion openstudio-gems definitely should not be packaging development dependencies... this was probably an oversight, although @macumber is probably the one who would know whether there was a reason not to exclude development dependencies.

PR #803, opened 2020-04-28
Correct surfaces 190, 515, and 516 in ASHRAELargeOfficeDetailed.osm

jugonzal07:
Hi Matt,

Large office detailed had a handful of small surfaces causing issues when testing the ground FC factor updates. Upon inspection, these surfaces had incorrect outside boundary conditions (see attached screenshot). They were incorrectly flagged as having an outside boundary condition to Ground. I corrected them in this pull request.

LargeOfficeDetailedBuggedSurfaces

### DIFF
--- a/data/geometry/ASHRAE9012013MediumOfficeDetailed.osm
+++ b/data/geometry/ASHRAE9012013MediumOfficeDetailed.osm
@@ -1,7 +1,7 @@
 
 OS:Version,
-  {27ab0958-a306-456b-8ced-d6a42410ac31}, !- Handle
-  2.3.0;                                  !- Version Identifier
+  {d5cdc5f7-d600-4f3d-a305-b30ddc0b9ace}, !- Handle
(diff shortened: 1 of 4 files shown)


jugonzal07:
To add some additional context to the screenshot I attached-- the circled surfaces are the ones I adjusted in the ASHRAELargeOfficeDetailed OSM file. They're level with the z=0 plane (i.e. they're "on grade") and floating above the basement floor you see below. The brown surfaces are those with outside boundary condition of "Ground", the rest are hidden.

I did not remove these problematic surfaces, only changed their outside boundary condition to Adiabatic like the floors beside them. I leave it to the owners of the repository to decide whether or not they should be exist in the ASHRAELargeOfficeDetailed geometry OSM file.

mdahlhausen:
The zones are overlapping in the geometry, causing the surface matching to fail.  I'll go ahead and fix the geometry.

PR #809, opened 2020-05-11
[90.1-AppG] Update WWR adjustment

lymereJ:
This PR updates the WWR adjustment for the Appendix G 90.1-2019 baseline model creation. It adjusts the WWR for all spaces of each building area type to match the requirements in G3.1.1-1 while adjusting it proportionally based on the proposed building window area distribution.

The unit test was updated to include the following steps:
- It now assigns a different standard space type to one of the space
- Run a sizing run to generate the SQL database
- Query the database for get the whole building adjusted WWR
- Compare it with the expected value 

The SRR adjustment method doesn't need to be adjusted as the rules for 90.1-2013 and 90.1-2019 are the same and are already implemented.

### DIFF
--- a/lib/openstudio-standards/standards/Standards.Model.rb
+++ b/lib/openstudio-standards/standards/Standards.Model.rb
@@ -3505,8 +3505,13 @@ class Standard
     space_cats = {}
 
     model.getSpaces.sort.each do |space|
-      # Get standards space type
-      std_spc_type = space.spaceType.get.standardsBuildingType.to_s
+      # Get standards space type and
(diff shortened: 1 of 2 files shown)


lymereJ:
As per this morning's discussion 3c9cf6f includes edits to the WWR adjustment method (and related test) so it only uses a single building area type (passed as an argument) to adjust the baseline model WWR. Note that if the building area type isn't passed to the method, it uses each space's space type's standands building type to make the adjustment, meaning that it still can handle multiple building area type as specified in 90.1-2019 Appendix G.

PR #852, opened 2020-07-06
[AppendixG PRM] handle lighting occupancy sensors

yunyangye:
Add the script to handle lighting occupancy sensors and modify the test json file accordingly.


yunyangye:
> Jeremy is right - the adjustment needs to be applied to the lighting schedules, not the LPD. So you need to look at the LIGHTS object to find the schedule name.

I see. It needs to create a new function to handle this. I will try to figure it out how to do it.

PR #879, opened 2020-08-06
support additional cooling equipment for heat pump loops

mdahlhausen:
This will need some minor object name changes in some prototype regression models (changed the name of cooling towers and fluid coolers to be more detailed).


mdahlhausen:
Not that we still don't have efficiency values for all cooling tower and fluid cooler types.
So this pull request is about enabling these systems for heat pump loops, but not setting the efficiency values per a standard.

PR #900, opened 2020-09-01
fix DCV application

mdahlhausen:
all thermal zones should be converted to oa per area *unless* DCV is required, not *if* DCV is required


mdahlhausen:
@lymereJ to update regression tests

lymereJ (approved):
Good catch! I'll push the updated results/regression models shortly.

PR #904, opened 2020-09-03
Tall and Supertall building prototype models

KaiyuSun:
Added the tall and supertall building prototype models, scorecard included too.

The only issue that remains to be fixed is the lack of OA node height in OpenStudio SDK. For a temporary fix, the LBNL team will develop an EnergyPlus measure and modify the current model development workflow in OS-Standards to allow the insertion of OS/E+ measures. This aims to be fixed before the next release (2021 March).


mdahlhausen (changes requested):
Finally got a chance to look at this.  The pull request has got all the piece, but there are some changes needed to make it less intrusive and not change tests for the PNNL DOE prototypes:

- revert the addition of the additional_params argument in model_custom_<x> methods.  There are other ways to handle the additions you want that are less intrusive to the code base.  Let's have a call to discuss if that would be helpful.
- any changes to Protyope.Model.HVAC should be universal - adding a number of chillers argument is too specific.  Ideally, this should be based on chiller size (e.g. above a certain size, split into two chillers). There is already logic in hvac_systems to do this.  Please remove that additional num_chillers property from the spreadsheet/jsons.
- When committing .json changes, only commit changes relevant to tall buildings.  That means manually going in and deleting the added college construction sets.  There isn't another way around this if we want to maintain open access of the spreadsheets.
- do not pszac fan types; the PNNL models should not be affected by your change
- do not limit tests for the doe_prototypes to fewer climate zones.

It may be easier to make a clean branch and just pull over the changes you need.

KaiyuSun:
@mdahlhausen Thank you Matthew for the review! Quick reply to some of your comments:
 
> * revert the addition of the additional_params argument in model_custom_ methods.  There are other ways to handle the additions you want that are less intrusive to the code base.  Let's have a call to discuss if that would be helpful.

Yes absolutely let's have a call! It is also closely related to the temporary fix method for the OA node height. I'll send you email separately to schedule it. 

> * any changes to Protyope.Model.HVAC should be universal - adding a number of chillers argument is too specific.  Ideally, this should be based on chiller size (e.g. above a certain size, split into two chillers). There is already logic in hvac_systems to do this.  Please remove that additional num_chillers property from the spreadsheet/jsons.

The "num_chillers" from **prototype_input** was already in the master branch, not actually added by me. I only added the part to assign num_chillers ONLY IF it is specified in **system_to_space_map** (i.e. the prototype building-specific hvac mapping json file). I then merged the logic of these two parts.
Besides, the logic of updating number of chillers currently only exists in "model_create_prm_baseline_building", but not in "model_create_prototype_model". Right now the number of chillers is by default always 1 in the prototype model except being specified. We can also chat about this during the call.

> * When committing .json changes, only commit changes relevant to tall buildings.  That means manually going in and deleting the added college construction sets.  There isn't another way around this if we want to maintain open access of the spreadsheets.

No problem. I'll remove the changes non-relevant to tall buildings.

> * do not pszac fan types; the PNNL models should not be affected by your change

Could you specify where the change is? I don't recall and couldn't locate the changes to fan type in psz-ac by searching through [code changes](https://github.com/NREL/openstudio-standards/pull/904/files).

> * do not limit tests for the doe_prototypes to fewer climate zones.

My bad. It was changed for debugging purposes. I'll reverse the changes.

> It may be easier to make a clean branch and just pull over the changes you need.

Good suggestion. Will do after our discussion.

mdahlhausen:
To do:

- remove adding additional parameters; we will figure out how to make the prototype mutable for a future release
- data .json files - remove the college changes and everything not tall buildings related.  Revert the space type .json changes, I will handle that
- rubocop
- revert changes to doe prototype tests

KaiyuSun:
@mdahlhausen I've updated the codes accordingly. Only thing left is the reversion of space type.json files (I've reverted other json files unrelated to tall buildings).

Besides, I didn't yet see a new build appear on the Amazon s3 bucket more than 24 hours after pushing. Any idea?

mdahlhausen:
Thanks @KaiyuSun 

When the branch is under a pull-request, the tests will show up under the PR number.  [Here are the tests for PR-904](http://cbci-build-logs.s3-website-us-east-1.amazonaws.com/?prefix=openstudio-standards/branches/PR-904/builds/2/).

KaiyuSun:
Gotcha! Thanks @mdahlhausen !

mdahlhausen (approved):
I made some very minor edits.  looks ready to merge.

PR #909, opened 2020-09-10
Update ashrae_90_1_2010.energy_recovery.json

mdahlhausen:
replace erroneous 0 values with nil in the energy recovery .json


mdahlhausen:
@lymereJ this should fix the regression errors for 1B, 2B, and 5C.

I suppose this is a good demonstration of why regression models beyond 4A would be helpful.

lymereJ (approved):

PR #911, opened 2020-09-15
fixed dcv types flags in necb_2011, autozone, and text_dcv

srgilani:
fixed dcv types flags (replaced spaces and hyphens with underscores) in necb_2011, autozone, and text_dcv


srgilani:
fixed dcv types flags (replaced spaces and hyphens with underscores) in necb_2011, autozone, and text_dcv

PR #919, opened 2020-09-18
Fixes OA conversion to per-area logic

asparke2:
For all AirLoopHVAC objects, a check is performed to determine whether any ThemalZone served requires DCV.  For a) zones that don't require DCV on airloops that do require DCV, and b) all zones on airloops that don't  require DCV, the OA requirements (per-person, per-area, and ACH) are combined and re-normalized into a single per-area number.  This is to prevent DCV actions from incorrectly acting on all zones on the airloop, and prevents unintentional interactions with other AirLoopHVAC controls requirements such as Multizone VAV Optimization.

Before this fix, because OA requirements were set by SpaceType, if one Space of a given SpaceType did require DCV but another Space of that same SpaceType did not require DCV, the OA requirement would incorrectly be converted for ALL Spaces of that SpaceType.  The result was that although DCV was "on," there were no per-person-based OA requirements to multiply by the real-time occupancy level, so DCV was not actually operating.

The fix was to create a new design OA object that mirrored the rates in the original OA object, and assign this new design OA object to the individual Spaces (as opposed to Space Types), so that inadvertent modifications are avoided.

If this sounds complex that's because it is.

(cherry picked from commit 529dbf8805c1206b1b60e862a210012560830189)


lymereJ (approved):

mdahlhausen:
@phylroy can you review this and regenerate NECB regression models?

mdahlhausen:
@ckirney I've updated the pull request with the changes I mentioned.  Ready for you to review.

PR #921, opened 2020-09-18
[Appendix G] Update the infiltration calculations

lymereJ:
After checking with the 90.1-2019 PRM development team regarding which space conditioning type should be used for the baseline model (we noticed that the space conditioning can change after assigning baseline internal load assumptions), it was decided to assume that the baseline space conditioning be the same as the proposed/user model.

A `TODO` was added as a reminder to take into account addendum db when adding rules for 90.1-2022 Appendix G.

The proposed changes also include a change in the default infiltration method from `Flow/Area` to `Flow/ExteriorWallArea` (more reasonable assumption). Changes also include a check to only count space floor area of non-unconditioned spaces that have non-zero surface area with outside boundaries set to outdoors (i.e. core spaces are excluded from the calculation since they have no infiltration, just transfer air).

The proposed changes also include an errata to the appendix language that by including floors in italic excluded the slab-on grade floors from the building envelope calculations.

### DIFF
--- a/lib/openstudio-standards/standards/Standards.Space.rb
+++ b/lib/openstudio-standards/standards/Standards.Space.rb
@@ -1460,6 +1460,7 @@ class Standard
         end
       end
     end
+
     return area_m2 * space.multiplier
   end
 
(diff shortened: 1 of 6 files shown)


lymereJ:
@dmaddoxwhite - I **think** that I covered most of what was brought up to Mike R. and Mike T., please have a look at the changes and let me know if you see any incorrect and/or if you think we should add additional tests. The test should pass. I'll merge it into the development branch after your review, and I'll then merge `apxg_hvac_type` so other have a "clean" starting point (some of the non-90.1-2019-PRM tests are still expected to fail).

dmaddoxwhite (approved):
Looks good!

PR #923, opened 2020-09-21
Baseline system mapping and assignment

lymereJ:
The proposed changes update the baseline system mapping and assignment following the 90.1-2019 PRM rules.


lymereJ:
@dmaddoxwhite - The appendix G test fails [here ](https://github.com/NREL/openstudio-standards/blob/b72dc2daf06e38eb5d628c0373ecafc847b7a358/lib/openstudio-standards/prototypes/common/objects/Prototype.Fan.rb#L135) because `Packaged_RTU_SZ_AC_CAV_OnOff_Fan` cannot be found on the line [above](https://github.com/NREL/openstudio-standards/blob/b72dc2daf06e38eb5d628c0373ecafc847b7a358/lib/openstudio-standards/prototypes/common/objects/Prototype.Fan.rb#L133). Do you have a more up-to-date version for the JSON files?

dmaddoxwhite:
It looks like you can use the version from lib\openstudio-standards\standards\ashrae_90_1\data to update the version in ashrae_90_1_prm\data.  This was the change that was made for the unitary system wrapper update.

lymereJ:
@dmaddoxwhite - The appendix G test still fails. Now, it's the infiltration test that fails. It fails because the space envelope area calculated by `space_envelope_area()` isn't the same as the expected values, it actually is really small. I tracked down the reason for it, and it is pretty convoluted: the baseline sizing run fails because the transformer added by #838 is associated to an EMS variable, however, EMS variables (and all EMS related objects) are deleted when creating the baseline model ([here](https://github.com/NREL/openstudio-standards/blob/abdf83a67866995b024fed61ff89b834e74f865a/lib/openstudio-standards/standards/Standards.Model.rb#L3808)) so the reference cannot be found. Since the sizing run fails, the heating/cooling loads that are used to determine space conditioning type cannot be found and most spaces are identified as unconditioned, which results in a low space envelope area. What's interesting is that even though the sizing run fails the baseline model is created anyway.

How should we tackle this issue? Below are a few thoughts:
1. Delete the transformers from the input prototypes
2. Edit `model_remove_prm_ems_objects()` to remove all objects associated with EMS variables

My preference would be for 2. as this seems to be the most clean way of doing it (sustainable), 1. would a really quick'n'dirty solution but could be an issue since transformer losses can be part of the baseline model (see G3.1.15).

dmaddoxwhite:
Since the sofware needs to work with any user model, approach #2 looks like the right one. Does "remove all objects" delete the transformer from the model?

lymereJ:
No, it doesn't. That's the main problem, it is still there but has broken references.

lymereJ:
We decided to go with 1. for now, we'll tackle the objects referencing deleted EMS objects later.

lymereJ:
@dmaddoxwhite - This changes should do it, at least it did it locally.

dmaddoxwhite (approved):

lymereJ:
@dmaddoxwhite - Looks like 55aad4f passed, but not 04db8bf, I probably missed something when merging.

lymereJ:
There were two methods that were misplaced. Tests are passing locally.

PR #926, opened 2020-09-21
Revert condenser loop changes

mdahlhausen:
@lymereJ there was a change to condenser water temperatures you committed back in May that I should have looked at closer before merging.  It is causing failures in other measures that just use openstudio-standards to build up HVAC systems.  The issue is that your method requires [a call to lookup the .ddy file](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb#L410-L418), which may not exist for a model.

The methods in `Prototype.hvac_systems` should be skeleton methods to add the system, and ideally control parameters adjusted later in `Standards.PlantLoop` method.  However, the control parameters for condenser loops influence sizing so they need to be set before the sizing run.  I debated whether to add this in the `model_add_cw_loop` method or call an addition hvac adjustment method before running sizing, and determined that most people would want the 90.1 default controls so it would be better to add them in the constructor.

I've removed your condenser loop code additions from `Prototype.hvac_systems` and instead add a `use_90_1_design_sizing` argument to the `model_add_cw_loop` method to by default override the other input arguments for system design temperatures and use the relevant 90.1 method.  [I do that here](https://github.com/NREL/openstudio-standards/blob/revert_condenser-loop-changes/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb#L529-L532), after the cooling tower constructor.

I'm using the `plant_loop_apply_prm_baseline_condenser_water_temperatures` method to set this, which was already in openstudio-standards.  This method looks up the design days in the model object, rather than loading and reading from the .ddy file.  I added the WB=>MDB conditions to the `model_add_design_days_and_weather_file` method.

There are two changes to the prototype models that will require regenerating regression models and performance results:
1) [The current version of the method](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb#L469) still uses the default wet bulb approach argument value in setting the offset in the condenser water setpoint manager.  This is used in design sizing and will change the cooling tower design flow rate and therefore fan sizing.  This should be aligned with [cooling tower design approach temperature here](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb#L530).
I've verified that when I run the LargeOffice 90.1-2004 regression test, I see a cooling tower sizing difference using the `plant_loop_apply_prm_baseline_condenser_water_temperatures` method, which sets the setpoint manager offset, and that there is no cooling tower sizing difference when I set the default wet bulb approach in the setpoint manager right after applying that method but before sizing.

2)  The prior method on master used the 90.1-2010 method in G3.1.3.11 to set the approach temperature.  This was actually changed in 90.1-2013 to a simpler formula, [which is defined here](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/standards/ashrae_90_1/ashrae_90_1_2013/ashrae_90_1_2013.PlantLoop.rb#L10-L40).  The `plant_loop_apply_prm_baseline_condenser_water_temperatures` method overloads the method for determining the approach temperature as shown in the code above. The 90.1-2013 Hospital and LargeOffice regression models haves issues from this and 1).

This pull request fixes both 1) and 2).

Please review the changes and if you approve, update the regression and performance models for LargeOffices and Hospitals.  Thanks!


mdahlhausen:
@DavidGoldwasser this should fix several measure test errors that involve condenser loops with cooling towers.

mdahlhausen:
@lymereJ ok, there is an interior lights error now in the regression tests.  I'll need to track down where that is coming from.

lymereJ:
@mdahlhausen - I reviewed the changes and talked to @jiangithub about it, we have a few comments.

- Thanks for catching the setpoint manager offset temperature difference! We did not make the change as part of #804, and frankly didn't notice that we should have, as we were reproducing changes that we had made to the Commercial Prototype Building models. These models do not use a `SetpointManager:FollowOutdoorAirTemperature` but use a `SetpointManager:Scheduled` instead which is probably not as efficient. This made us realize that we'll have to update the code base so that 90.1 openstudio-standards prototypes are aligned with the Commercial Prototype Building models.
- We understand the issue of having `.ddy` files used in a skeleton method that is potentially used by other measures. I think we both agree that we still need to tie the model creation to design weather data. We are not sure if the method you propose (adding a `SizingPeriod:DesignDay` object) is appropriate as it impacts sizing of autosized equipment. We confirmed that by creating and running simulation for prototypes with [these ](https://github.com/NREL/openstudio-standards/pull/926/files#diff-b79b6093deef5ef146492297d8efff89) changes alone and without, the WB=>MDB design day tends to drive the capacity of equipment up. We are not opposed to adding `SizingPeriod:DesignDay` objects (we are for example looking into adding multiple cooling design days based DB=>MWB in the Commercial Prototype Building models) but if it justifiably impact sizing of the equipment, this is doesn't seem to be the case when adding a WB=>MDB cooling design day.
- We actually intended to perform sizing base on the 90.1-2010 approach for all versions of 90.1. The Commercial Prototype Building model assumptions are based on prescriptive code requirements. There has been a code change proposal to include the approach formula from 90.1-2013 to the prescriptive requirements in 90.1 but it failed and it was only added to the PRM so all 90.1 openstudio-standards prototypes models should follow the 90.1-2010 approach. We are not opposed to using the existing prm code that you propose (`plant_loop_apply_prm_baseline_condenser_water_temperatures`) to use. The issue being that it is tied to the version of 90.1 (`plant_loop_prm_baseline_condenser_water_temperatures`), and so currently the 90.1-2013 prototypes are sized using the 90.1-2013 PRM approach which is not appropriate (for the prototype building model). While `plant_loop_prm_baseline_condenser_water_temperatures` in `Standards.PlantLoop.rb` makes mostly the same changes as #804, it doesn't seem to take into account the EnergyPlus cooling tower algorithm limits (1.11deg. C and 11.11 deg. C for the approach, 20 deg. C and 26.67 deg. C for the inlet air WB temperature). Perhaps this is what you meant by overloads.

lymereJ (changes requested):
@mdahlhausen - I reviewed the changes and talked to @jiangithub about it, we have a few comments.

- Thanks for catching the setpoint manager offset temperature difference! We did not make the change as part of #804, and frankly didn't notice that we should have, as we were reproducing changes that we had made to the Commercial Prototype Building models. These models do not use a `SetpointManager:FollowOutdoorAirTemperature` but use a `SetpointManager:Scheduled` instead which is probably not as efficient. This made us realize that we'll have to update the code base so that 90.1 openstudio-standards prototypes are aligned with the Commercial Prototype Building models.
- We understand the issue of having `.ddy` files used in a skeleton method that is potentially used by other measures. I think we both agree that we still need to tie the model creation to design weather data. We are not sure if the method you propose (adding a `SizingPeriod:DesignDay` object) is appropriate as it impacts sizing of autosized equipment. We confirmed that by creating and running simulation for prototypes with [these ](https://github.com/NREL/openstudio-standards/pull/926/files#diff-b79b6093deef5ef146492297d8efff89) changes alone and without, the WB=>MDB design day tends to drive the capacity of equipment up. We are not opposed to adding `SizingPeriod:DesignDay` objects (we are for example looking into adding multiple cooling design days based DB=>MWB in the Commercial Prototype Building models) but if it justifiably impacts sizing of the equipment, this is doesn't seem to be the case when adding a WB=>MDB cooling design day.
- We actually intended to perform sizing base on the 90.1-2010 approach for all versions of 90.1. The Commercial Prototype Building model assumptions are based on prescriptive code requirements. There has been a code change proposal to include the approach formula from 90.1-2013 to the prescriptive requirements in 90.1 but it failed and it was only added to the PRM so all 90.1 openstudio-standards prototypes models should follow the 90.1-2010 approach. We are not opposed to using the existing prm code that you propose (`plant_loop_apply_prm_baseline_condenser_water_temperatures`) to use. The issue being that it is tied to the version of 90.1 (`plant_loop_prm_baseline_condenser_water_temperatures`), and so currently the 90.1-2013 prototypes are sized using the 90.1-2013 PRM approach which is not appropriate (for the prototype building model). While `plant_loop_prm_baseline_condenser_water_temperatures` in `Standards.PlantLoop.rb` makes mostly the same changes as #804, it doesn't seem to take into account the EnergyPlus cooling tower algorithm limits (1.11deg. C and 11.11 deg. C for the approach, 20 deg. C and 26.67 deg. C for the inlet air WB temperature). Perhaps this is what you meant by overloads.

mdahlhausen:
@lymereJ 
I reverted the addition of design days and copied the condenser water loop temperature sizing objects to a new method in a new Prototype.CoolingTower file using the 90.1-2010 approach.  This addresses your last two points.  The design cooling tower sizing change is still there from your first point, and I'll need you to update the regression models accordingly (LargeOffice, Hospital).

lymereJ (approved):
Looks good! No significant IDF changes other than setpoint related when looking at the full set of prototypes.

PR #945, opened 2020-10-07
set fan operating model schedules unitary equipment objects when ventilation present

mdahlhausen:
This prevents unitary system objects from running always on and instead following the zone occupancy schedule, likewise for zone hvac equipment objects if they provide ventilation
- for air loops, adjusts the supply fan operating mode schedule to follow the loop occupancy schedule (typically a single zone) for AirLoopHVACUnitarySystem objects
- for zone hvac systems, sets the supply fan operating mode schedule to the zone occupancy schedule if ventilation; otherwise set it to always off to follow the load.  This represents a change for PTAC and PTHP systems, which were always on


lymereJ:
Looks like we can't have a `Fan:ConstantVolume` object with a fan operation schedule that has values set to 0 when using a `ZoneHVAC:PackagedTerminalAirConditioner`. The small hotel prototype models (from the Commercial Prototype models set) uses a `Fan:OnOff` with a schedule that's always on. In 62.2 ventilation can be both intermittent or continuous. Continuous ventilation has been chosen. The proposed changes needs to be adjusted to take this into account. We can propose some modifications if you would like.

lymereJ (changes requested):
The proposed changes set the `ZoneHVAC:PackagedTerminalAirConditioner` object in the small hotel with `Fan:ConstantVolume` objects that have an always off fan operating which crashes the sizing runs. The small hotel prototype models (from the Commercial Prototype models set) uses a `Fan:OnOff` with a schedule that's always on. In 62.2 ventilation can be both intermittent or continuous. Continuous ventilation was chosen. The proposed changes needs to be adjusted to take this into account. We can propose some modifications if you would like.

mdahlhausen:
@lymereJ yep, saw those issues too.
I made another commit making all unitary system fans OnOff instead of Constant Volume.
This should resolve that issue.  I'll walk you through the changes tomorrow AM.

lymereJ (approved):

PR #948, opened 2020-10-09
Apxg add sz cv

dmaddoxwhite:
Add new baseline system type for 2019 Appendix G: constant volume four pipe fan coil.


lymereJ (approved):
Looks great! The usage for the 4-pipe fan coil is clever.

HillJamison1:
unsubscribe

PR #951, opened 2020-10-10
Update SRR adjustment

lymereJ:
The proposed changes add a new 90.1-PRM method to adjust the skylight-to-roof ratio of the baseline model. The existing method was based on the 90.1-2010 PRM user manual which is different from the 90.1-2019 PRM. A "SRR" test was also added.


dmaddoxwhite (approved):
Looks good!

PR #957, opened 2020-10-19
Nrcan 275 pv

srgilani:


lib/openstudio-standards/standards/necb/ECMS/pv_ground.rb
```diff
@@ -0,0 +1,57 @@
+class ECMS
+
+  def apply_pv_ground(model:, pv_ground_type:, pv_ground_total_area_pv_panels_m2:, pv_ground_tilt_angle:, pv_ground_azimuth_angle:, pv_ground_module_description:)
+
+    ##### If pv_type is nil.. do nothing.
+    return if pv_ground_type.nil? || pv_ground_type == FALSE
+
+    ##### Set default PV panels' tilt angle as the latitude
+    if pv_ground_tilt_angle == 'NECB_Default'
```
phylroy, line 9: if pv_ground_tilt_angle is nil..or undefined... what will happen here?

phylroy: add similar nil/false here to allow setting of default value without having to set the switch. Similar to what you did above.

phylroy (changes requested):

lib/openstudio-standards/standards/necb/ECMS/pv_ground.rb
srgilani: Added below lines for "do nothing" option in lib/openstudio-standards/standards/necb/ECMS/pv_ground.rb:
   return if pv_ground_total_area_pv_panels_m2.nil? || pv_ground_total_area_pv_panels_m2 == FALSE
   return if pv_ground_tilt_angle.nil? || pv_ground_tilt_angle == FALSE
   return if pv_ground_azimuth_angle.nil? || pv_ground_azimuth_angle == FALSE
   return if pv_ground_module_description.nil? || pv_ground_module_description == FALSE

PR #967, opened 2020-11-18
Add system 11 and related SAT controls

lymereJ:
Add baseline system 11 (single-zone vav for computer rooms and datacenters) and related SAT controls. Also fix formatting offenses (according to Rubocop) of the appendix g test file.


lymereJ:
@dmaddoxwhite - Please, let me know what you think of the changes and especially of the new tests. I could definitely expand on them (for example checking the heating coil based on climate zone for system 11). The modification to the JSON files have not yet been made to the Google/Excel sheet. I'll wait for your review in case changes are needed.

Non-SAT test changes were made to address rubocop offenses.

dmaddoxwhite:
I think the fan curve for the new fan type in OpenStudio_Standards_fans.json should be "Multi Zone VAV with VSD and fixed SP setpoint", based on G3.1.3.15.

lymereJ:
The fans in `OpenStudio_Standards_fans.json` are not used for the 2019 PRM. The new fan entry in `ashrae_90_1_prm/data/ashrae_90_1_prm.fans.json` has the fan curves from G3.1.3.15.

dmaddoxwhite:
OK, I hadn't reviewed those file yet.

dmaddoxwhite:
In check_sat_ctrl(), you set spm_check = false, and then reset to true if any spm_check = 5. Should it instead initialize to true and reset to false if any check fails?

dmaddoxwhite:
Also, here is a list of computer room checks that I was planning to implement. I think #1 and #5 can be done with an unmodified OfficeLarge prototype.  I think getting this full list of possibilities for system type is more important that testing for heating type, since heating will not likely get used.

1. Use OfficeLarge to check if small computer rooms with less than 600 kBtuh are PSZ.
2. Use OfficeMedium, and add a computer room with > 600 kBtuh and < 3 MMBtuh, and check if PSZ-AC is used for the computer room.
3. Use OfficeMedium, and add a computer room with > 3 MMBtuh, and check if System 11 with chiller is used, even though the overall system for OfficeMedium is PVAV.
4. Use OfficeLarge with a computer room with load between 600 kBtuh and 3 MMBtuh; check if System 11 is used.
5. Use OfficeLarge with a computer room with load > 3 MMBtuh; check if System 11 is used.

lymereJ:
> In check_sat_ctrl(), you set spm_check = false, and then reset to true if any spm_check = 5. Should it instead initialize to true and reset to false if any check fails?

I believe that what you are proposing and what's in this PR is equivalent:
1. What you are proposing is that by default the case always pass and it fails if the SPM:W and a dT of 5F isn't in the model.
2. What's in the PR assumes that by default the case always fails and it only pass if a SPM:W is used with a dT of 5F.

I have a preference for what is in the PR (approach 2.) because the test will fail by default and only passes if the check is correct whereas in 1., if there is something wrong with the check, the test will pass. I can revise if you think the approach 1. is better.

> Also, here is a list of computer room checks that I was planning to implement. I think #1 and #5 can be done with an unmodified OfficeLarge prototype. I think getting this full list of possibilities for system type is more important that testing for heating type, since heating will not likely get used.
> 
> 1. Use OfficeLarge to check if small computer rooms with less than 600 kBtuh are PSZ.
> 2. Use OfficeMedium, and add a computer room with > 600 kBtuh and < 3 MMBtuh, and check if PSZ-AC is used for the computer room.
> 3. Use OfficeMedium, and add a computer room with > 3 MMBtuh, and check if System 11 with chiller is used, even though the overall system for OfficeMedium is PVAV.
> 4. Use OfficeLarge with a computer room with load between 600 kBtuh and 3 MMBtuh; check if System 11 is used.
> 5. Use OfficeLarge with a computer room with load > 3 MMBtuh; check if System 11 is used.

Sounds good, I'll add those cases.

dmaddoxwhite:
Regarding the setpoint manager delta T check, the existing logic first sets spm_check = false, then loops through the setpoint managers. If any setpoint manager passes the test, then spm_check will be true, even though some of them might have failed.  Does that give a false report, or am I not understanding the logic of the test?

lymereJ:
I think I see your point now.

EnergyPlus won't respond well to having multiple SPMs on a single node (executed in order), the check was written with that in mind, meaning that if there's a SPM on the loop it's either a SPM:W or another. The OpenStudio app won't let you add multiple SPMs to a node for this reason but it looks like the SDK allows you to add multiple SPM to a single node.

I will had a guard clause that fails the check if there's multiple SPMs on the supply air loop outlet node, as it really should be checked for so `openstudio-standards` doesn't end-up adding multiple SPMs on single nodes. I see why you proposed approach 1. but I feel like it should be a dedicated assertion (hence the guard clause) so it's easier to see why the test fails. Perhaps it might actually be better to completely decouple the SAT check and number of SPM per node, and have a more general and dedicated check for the number of SPMs per node which would look at all possible air loop nodes and make sure that only one SPM is used on each node. What do you think?

dmaddoxwhite:
It seems like it makes sense to put the check here, since it is connected to the logic of this test. Since we do not generally expect there to be more than one SPM, then having a dedicated test for that might not make sense.

lymereJ:
Sounds good. That's what I added in the latest commit, I also added additional test cases to cover the list you provided. I added additional tests that check for the building computer room load instead of individual computer room test.

dmaddoxwhite (commented):
Looks good!

PR #985, opened 2020-12-16
Code which prevents the addition of air loops depending on the state of a switch.

ckirney:


ckirney:
Will use use ECMS class instead to remove code air loops rather than prevent them from being created.

PR #994, opened 2020-12-30
Data library export

mdahlhausen:
This pull requests adds a rake task "update:export:jsons" which exports 90.1 json data to a separate directory within the data folder, to be used for a separate data library.

This work wholly replicates and replaces the work on the standards-separation branch.


mdahlhausen:
@jiangithub and @lymereJ I can talk about this on our next call.  Basically, this exports out most of the 90.1 .json content.  We can review and decide what should be excluded if you'd like.

PR #1009, opened 2021-01-25
Use AirLoopHVAC:UnitarySystem for PSZ systems

yunyangye:
Some situations for psz-ac did not use unitary system before (airloop only). To align with other situations with unitary system, this PR modifies the script about psz-ac and makes all situations use unitary system. In addition, there are some duplicate scripts in the previous version. Thus, the script is simplified.

### DIFF
--- a/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
+++ b/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
@@ -2575,50 +2575,49 @@ class Standard
       unitary_system.setSupplyFan(fan) unless fan.nil?
       unitary_system.setHeatingCoil(htg_coil) unless htg_coil.nil?
       unitary_system.setCoolingCoil(clg_coil) unless clg_coil.nil?
+      unitary_system.setControllingZoneorThermostatLocation(zone)
+      unitary_system.setFanPlacement(fan_location)
+      unitary_system.setAvailabilitySchedule(model.alwaysOnDiscreteSchedule)
+      unitary_system.addToNode(air_loop.supplyInletNode)
       if fan_type == 'Cycling'
-        unitary_system.setControllingZoneorThermostatLocation(zone)
         unitary_system.setSupplyAirFanOperatingModeSchedule(model.alwaysOffDiscreteSchedule)
-        if heating_type == 'Water To Air Heat Pump' || heating_type == 'Single Speed Heat Pump'
+        case heating_type
+        when 'Water To Air Heat Pump'
           unitary_system.setSupplementalHeatingCoil(supplemental_htg_coil) unless supplemental_htg_coil.nil?
-        end
-        if heating_type == 'Water To Air Heat Pump'
           unitary_system.setName("#{zone.name} Unitary HP")
           unitary_system.setMaximumSupplyAirTemperature(dsgn_temps['zn_htg_dsgn_sup_air_temp_c'])
           unitary_system.setSupplyAirFlowRateMethodDuringCoolingOperation('SupplyAirFlowRate')
           unitary_system.setSupplyAirFlowRateMethodDuringHeatingOperation('SupplyAirFlowRate')
           unitary_system.setSupplyAirFlowRateMethodWhenNoCoolingorHeatingisRequired('SupplyAirFlowRate')
-        elsif heating_type == 'Single Speed Heat Pump'
-          unitary_system.setAvailabilitySchedule(model.alwaysOnDiscreteSchedule)
+        when 'Single Speed Heat Pump'
+          unitary_system.setSupplementalHeatingCoil(supplemental_htg_coil) unless supplemental_htg_coil.nil?
           unitary_system.setName("#{air_loop.name} Unitary HP")
           unitary_system.setMaximumOutdoorDryBulbTemperatureforSupplementalHeaterOperation(OpenStudio.convert(40.0, 'F', 'C').get)
         else
-          unitary_system.setAvailabilitySchedule(model.alwaysOnDiscreteSchedule)
           unitary_system.setName("#{air_loop.name} Unitary System")
         end
       else
-        unitary_system.setAvailabilitySchedule(model.alwaysOnDiscreteSchedule)
-        unitary_system.setSupplementalHeatingCoil(supplemental_htg_coil) unless supplemental_htg_coil.nil?
-        if heating_type == 'Single Speed Heat Pump'
+        unitary_system.setSupplyAirFanOperatingModeSchedule(hvac_op_sch)
+        unitary_system.setName("#{air_loop.name} Unitary AC")
+        case heating_type
+        when 'Single Speed Heat Pump'
+          unitary_system.setSupplementalHeatingCoil(supplemental_htg_coil) unless supplemental_htg_coil.nil?
           unitary_system.setMaximumOutdoorDryBulbTemperatureforSupplementalHeaterOperation(OpenStudio.convert(40.0, 'F', 'C').get)
           unitary_system.setName("#{air_loop.name} Unitary HP")
-        elsif fan_location == 'DrawThrough'
+        when 'Water'
+          unitary_system.setSupplementalHeatingCoil(supplemental_htg_coil) unless supplemental_htg_coil.nil?
           unless htg_coil.nil?
             # if water coil, rename controller b/c it is recreated when added to node
-            htg_coil.controllerWaterCoil.get.setName("#{htg_coil.name} Controller") if heating_type == 'Water'
+            htg_coil.controllerWaterCoil.get.setName("#{htg_coil.name} Controller") if fan_location == 'DrawThrough'
           end
           unless clg_coil.nil?
             # if water coil, rename controller b/c it is recreated when added to node
-            clg_coil.controllerWaterCoil.get.setName("#{clg_coil.name} Controller") if cooling_type == 'Water'
+            clg_coil.controllerWaterCoil.get.setName("#{clg_coil.name} Controller") if fan_location == 'DrawThrough'
           end
-          unitary_system.setName("#{air_loop.name} Unitary AC")
         else
-          unitary_system.setName("#{air_loop.name} Unitary AC")
+          unitary_system.setName("#{air_loop.name} Unitary System")
         end
-        unitary_system.setControllingZoneorThermostatLocation(zone)
-        unitary_system.setSupplyAirFanOperatingModeSchedule(hvac_op_sch)
       end
-      unitary_system.setFanPlacement(fan_location)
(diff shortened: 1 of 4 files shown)


lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2551,111 +2551,74 @@ def model_add_psz_ac(model,
...
+        return false
+      end
+
+      # wrap coils in a unitary system
+      unitary_system = OpenStudio::Model::AirLoopHVACUnitarySystem.new(model)
+      unitary_system.setSupplyFan(fan) unless fan.nil?
+      unitary_system.setHeatingCoil(htg_coil) unless htg_coil.nil?
+      unitary_system.setCoolingCoil(clg_coil) unless clg_coil.nil?
+      if fan_type == 'Cycling'
+        unitary_system.setControllingZoneorThermostatLocation(zone)
```
lymereJ, line 2579: This can be taken out of the if statement as it applies to any value of `fan_type`.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2551,111 +2551,74 @@ def model_add_psz_ac(model,
...
-          clg_coil.addToNode(air_loop.supplyInletNode) unless clg_coil.nil?
-          htg_coil.addToNode(air_loop.supplyInletNode) unless htg_coil.nil?
-          fan.addToNode(air_loop.supplyInletNode) unless fan.nil?
+          unitary_system.setName("#{air_loop.name} Unitary AC")
         else
-          OpenStudio.logFree(OpenStudio::Error, 'openstudio.model.Model', 'Invalid fan location')
-          return false
+          unitary_system.setName("#{air_loop.name} Unitary AC")
         end
+        unitary_system.setControllingZoneorThermostatLocation(zone)
```
lymereJ, line 2617: This can be taken out of the if statement as it applies to any value of `fan_type`.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2551,111 +2551,74 @@ def model_add_psz_ac(model,
...
+        unitary_system.setSupplyAirFanOperatingModeSchedule(hvac_op_sch)
       end
+      unitary_system.setFanPlacement(fan_location)
+      unitary_system.addToNode(air_loop.supplyInletNode)
```
lymereJ, line 2621: Can we move these lines below 2577 since it applies to all situations? It would be easier to read.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2551,111 +2551,74 @@ def model_add_psz_ac(model,
...
-          # Use a Fan:ConstantVolume placed directly on the air loop
-          fan = create_fan_by_name(model,
-                                   'Packaged_RTU_SZ_AC_CAV_Fan',
-                                   fan_name: "#{air_loop.name} Fan")
-          fan.setAvailabilitySchedule(hvac_op_sch)
-          supplemental_htg_coil.addToNode(air_loop.supplyInletNode) unless supplemental_htg_coil.nil?
-          clg_coil.addToNode(air_loop.supplyInletNode) unless clg_coil.nil?
-          htg_coil.addToNode(air_loop.supplyInletNode) unless htg_coil.nil?
-          fan.addToNode(air_loop.supplyInletNode) unless fan.nil?
+          unitary_system.setName("#{air_loop.name} Unitary AC")
```
lymereJ, line 2613: This line is the same as 2615 but they do the same thing. We should only have one of them.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2551,111 +2551,74 @@ def model_add_psz_ac(model,
...
+
+      # wrap coils in a unitary system
+      unitary_system = OpenStudio::Model::AirLoopHVACUnitarySystem.new(model)
+      unitary_system.setSupplyFan(fan) unless fan.nil?
+      unitary_system.setHeatingCoil(htg_coil) unless htg_coil.nil?
+      unitary_system.setCoolingCoil(clg_coil) unless clg_coil.nil?
+      if fan_type == 'Cycling'
+        unitary_system.setControllingZoneorThermostatLocation(zone)
+        unitary_system.setSupplyAirFanOperatingModeSchedule(model.alwaysOffDiscreteSchedule)
+        if heating_type == 'Water To Air Heat Pump' || heating_type == 'Single Speed Heat Pump'
```
lymereJ, line 2581: Could we use `case` statements (one for heating coil type) instead of `if`s? As it stands it's a bit difficult to review as we consecutive `if`s.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2551,111 +2551,74 @@ def model_add_psz_ac(model,
...
-          fan.setAvailabilitySchedule(hvac_op_sch)
-          supplemental_htg_coil.addToNode(air_loop.supplyInletNode) unless supplemental_htg_coil.nil?
-          clg_coil.addToNode(air_loop.supplyInletNode) unless clg_coil.nil?
-          htg_coil.addToNode(air_loop.supplyInletNode) unless htg_coil.nil?
-          fan.addToNode(air_loop.supplyInletNode) unless fan.nil?
+          unitary_system.setName("#{air_loop.name} Unitary AC")
         else
-          OpenStudio.logFree(OpenStudio::Error, 'openstudio.model.Model', 'Invalid fan location')
-          return false
+          unitary_system.setName("#{air_loop.name} Unitary AC")
```
lymereJ, line 2615: Shouldn't the name be `Unitary System` to be consistent with line 2596?

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2551,111 +2551,74 @@ def model_add_psz_ac(model,
...
           unitary_system.setName("#{air_loop.name} Unitary HP")
-          unitary_system.setControllingZoneorThermostatLocation(zone)
           unitary_system.setMaximumOutdoorDryBulbTemperatureforSupplementalHeaterOperation(OpenStudio.convert(40.0, 'F', 'C').get)
-          unitary_system.setFanPlacement(fan_location)
-          unitary_system.setSupplyAirFanOperatingModeSchedule(model.alwaysOffDiscreteSchedule)
-          unitary_system.addToNode(air_loop.supplyInletNode)
         else
-          # heating_type = NaturalGas, Electricity, Water or nil (no heat)
-          unitary_system = OpenStudio::Model::AirLoopHVACUnitarySystem.new(model)
           unitary_system.setAvailabilitySchedule(model.alwaysOnDiscreteSchedule)
```
lymereJ, line 2595: This should probably be taken out of the if statement as it applies to all configuration of unitary systems.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2551,111 +2551,74 @@ def model_add_psz_ac(model,
...
           unitary_system.setName("#{air_loop.name} Unitary System")
-          unitary_system.setControllingZoneorThermostatLocation(zone)
-          unitary_system.setFanPlacement(fan_location)
-          unitary_system.setSupplyAirFanOperatingModeSchedule(model.alwaysOffDiscreteSchedule)
-          unitary_system.addToNode(air_loop.supplyInletNode)
         end
       else
-        # ConstantVolume: Packaged Rooftop Single Zone Air conditioner
-        # Need unitary system wrapper for heat pumps in order to allow control of supplemental heat
+        unitary_system.setAvailabilitySchedule(model.alwaysOnDiscreteSchedule)
```
lymereJ, line 2599: This should probably be taken out of the if statement as it applies to all configuration of unitary systems.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2551,111 +2551,74 @@ def model_add_psz_ac(model,
...
           unless htg_coil.nil?
-            htg_coil.addToNode(air_loop.supplyInletNode)
             # if water coil, rename controller b/c it is recreated when added to node
             htg_coil.controllerWaterCoil.get.setName("#{htg_coil.name} Controller") if heating_type == 'Water'
           end
           unless clg_coil.nil?
-            clg_coil.addToNode(air_loop.supplyInletNode)
             # if water coil, rename controller b/c it is recreated when added to node
             clg_coil.controllerWaterCoil.get.setName("#{clg_coil.name} Controller") if cooling_type == 'Water'
           end
```
lymereJ, line 2612: Do you know why this is depend on `fan_location`? Shouldn't it only be dependent on `heating_type`?

lymereJ (changes requested):
Looks good! I think we just need to move things around a bit to make it more clear.

lymereJ (approved):

lymereJ:
@mdahlhausen - We noticed that there's 90.1-2013 PRM test failure, which seems expected since we're now using unitary systems for PSZ systems. You mentioned updating/adding some of the tests, we're not sure if you were planning on updating this one.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2551,109 +2551,71 @@ def model_add_psz_ac(model,
         clg_coil = nil
       end
 
-      # wrap coils in a unitary system if cycling
+      # cycling fan
       if fan_type == 'Cycling'
```
mdahlhausen, line 2555: this if / else statement is a mess.  It include fan_type and fan_location.  these should be independent

mdahlhausen:
Moving to #1019

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
yunyangye: Last time, I kept the if/else statement like the previous version, but we agree that it is better to avoid a mess; so, I address your comment and commit the branch. 
There are still two questions:
(1) in Lines 2613 and 2617, there are still messes. Should we move these two lines out or keep the current way?
(2) Like Jeremy mentioned before, we noticed that there's a 90.1-2013 PRM test failure, which seems expected since we're now using unitary systems for PSZ systems. You mentioned updating/adding some of the tests, we're not sure if you were planning on updating this one.
Thanks.

Issue #893, opened 2020-08-27
refactor add hvac tests


mdahlhausen:
The default model is in climate zone 7A which causes problems for cooling only systems (direct evap), VAV no reheat.  Expand the inputs to optionally override which model and climate zone to test in.

mdahlhausen:
Addressed with #1011

PR #1011, opened 2021-01-28
Refactor hvac tests

mdahlhausen:
- move it to a separate directory, os_stds_methods in the test directory
- break out tests into separate test for each system type
- do a sizing run and apply system efficiencies and controls
- allow additional variation in climate zone
- add logic to model_add_hvac system to default arguments if inputs are nil
- have model_apply_hvac_efficiency_standard return true if it completes successfully


mdahlhausen:
@asparke2 when you get a chance, here is a major refactor of the hvac system tests
I'm running it now on the test machine and will likely need to make changes and will want to modify tests, but the structure will be the same
Is there any other variation you think we should include to get broader coverage?
@lymereJ that same question applies - any system/climate/building type variations you think we should apply?

lymereJ:
* The system type selection seems pretty exhaustive
* Perhaps we could include the following building types as well: school (primary?), hospital, retail (standalone?), warehouse, and hotel (large?)
* It might also be good to include 2B, 3B, 4A, 4B, 4C, and perhaps 8 instead of 7?

Let me know if you think this is too much!

mdahlhausen:
@asparke2 this is ready to merge

@lymereJ I think either office or school and hotel or multifamily, plus hospital is good coverage.  For climate zones, 2B and 7 should capture most conditions.  I'll add these in the next round.

mdahlhausen:
Addresses #893

mdahlhausen:
leaving the branch open to add additional test support

PR #1014, opened 2021-02-05
Fluid coolers in Large Office Prototype's Data Centers

jugonzal07:
Hi Matt,

The data centers in the Large Office Prototype were incorrectly utilizing cooling towers when they should be using fluid coolers instead. This change addresses this problem. 

On a high level, these are changes I made:

- Updated `ASHRAELargeOffice.json` to specify the `heat_pump_loop_cooling_type` as 'FluidCoolerTwoSpeed'
- Updated `Prototype.hvac_systems.rb` to set default fluid cooler minimum power per flow when creating FluidCoolerTwoSpeed objects. This was done using an existing OSS function.
- Updated both the Google Drive Spreadsheet as well as the 90.1 heat rejection jsons to account for minimum performance metrics for FluidCoolerTwoSpeed objects.

We can update the test results once you've reviewed the changes.


lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -749,6 +749,7 @@ def model_add_hp_loop(model,
       when 'FluidCoolerTwoSpeed'
         cooling_equipment = OpenStudio::Model::FluidCoolerTwoSpeed.new(model)
         cooling_equipment.setName("#{heat_pump_water_loop.name} FluidCoolerTwoSpeed")
+        fluid_cooler_apply_minimum_power_per_flow(cooling_equipment)
```
mdahlhausen, line 752: Applying efficiency standards should not be a part of this part of code.  It should instead be called here:
https://github.com/NREL/openstudio-standards/blob/large_office_fluid_cooler/lib/openstudio-standards/standards/Standards.Model.rb#L1545-L1549

lib/openstudio-standards/standards/Standards.FluidCooler.rb
```diff
@@ -36,7 +36,7 @@ def fluid_cooler_apply_minimum_power_per_flow(fluid_cooler)
     search_criteria['template'] = template
 
     # Closed cooling towers are fluidcooler objects.
-    search_criteria['equipment_type'] = 'Closed Cooling Tower'
+    search_criteria['equipment_type'] = 'Dry Cooler'
```
mdahlhausen, line 39: Perhaps this needs to be an argument?  Per 90.1-2013 Table 6.8.1-7, there are values for '...closed-circuit cooling towers' and 'air cooled condensers'.  Maybe this is an additional property for the fluid cooler object?

mdahlhausen (changes requested):
Some minor comments - we need to distinguish between closed-circuit and dry coolers, and move the efficiency application to the model_apply_hvac_efficiency_standard method (where it is now commented out)

mdahlhausen:
@jugonzal07 @lymereJ this is almost ready to merge.  I added some documentation, and extended it to the ze aedg and nze templates.  The secondary school model is failing the regression test on the last commit.  Can you review after the current CI test is finished and resubmit a new regression model for that?

mdahlhausen (approved):

PR #1016, opened 2021-02-15
Revised ground-mounted PV measure

srgilani:
revised pv measure and necb2011 to consider building footprint as default area for pv panels, and set necb_default for pv_ground_type as do-nothing for pv measure, and added none option for all args of pv measure

### DIFF
--- a/lib/openstudio-standards/standards/necb/ECMS/pv_ground.rb
+++ b/lib/openstudio-standards/standards/necb/ECMS/pv_ground.rb
@@ -2,12 +2,34 @@ class ECMS
 
   def apply_pv_ground(model:, pv_ground_type:, pv_ground_total_area_pv_panels_m2:, pv_ground_tilt_angle:, pv_ground_azimuth_angle:, pv_ground_module_description:)
 
+    ##### Remove leading or trailing whitespace in case users add them in inputs
+    if pv_ground_total_area_pv_panels_m2.instance_of?(String)
+      pv_ground_total_area_pv_panels_m2 = pv_ground_total_area_pv_panels_m2.strip
+    end
+    if pv_ground_tilt_angle.instance_of?(String)
+      pv_ground_tilt_angle = pv_ground_tilt_angle.strip
+    end
+    if pv_ground_azimuth_angle.instance_of?(String)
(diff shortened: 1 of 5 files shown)


test/necb/unit_tests/expected_results/pv_ground_expected_results.json
```diff
@@ -5,7 +5,7 @@
     "building_type": "FullServiceRestaurant",
     "primary_heating_fuel": "DefaultFuel",
     "pv_ground_type": true,
-    "Generator PVWatts 1 - dc_system_capacity": "8611.12833336778",
```
phylroy, line 8: wow big change.. was this a correction from mike?

lib/openstudio-standards/standards/necb/ECMS/pv_ground.rb
```diff
@@ -2,12 +2,52 @@ class ECMS
...
+      pv_ground_azimuth_angle = pv_ground_azimuth_angle.to_f
+    end
+
+    ##### Calculate footprint of the building model (this is used as default value for pv_ground_total_area_pv_panels_m2)
+    building_footprint_m2 = 0.0
+    model.getSpaces.sort.each do |space|
+      space.surfaces.sort.each do |surface|
+        if surface.surfaceType == 'Floor' && surface.outsideBoundaryCondition == 'Ground'
+          floor_vertices = surface.vertices
+          if floor_vertices[0].z == 0.0
```
phylroy, line 40: What would happen if the ground floor was not at exactly zero?

phylroy (changes requested):

lib/openstudio-standards/standards/necb/ECMS/pv_ground.rb
srgilani: Revised to the lowest floors' area (assuming the floor of all lowest spaces are at the same level).

test/necb/unit_tests/expected_results/pv_ground_expected_results.json
srgilani: No, previously I set the default area of the PV panels as 50 m2, whereas I set it to a building's footprint in my revisions.

phylroy (approved):

PR #1017, opened 2021-02-16
Heat rejection updates

jugonzal07:
Hi Jeremy,

I have updated heat rejection performance metrics for 90.1 2016/2019 based on Table 6.8.1-7.  This also includes updated default performance for Dry Coolers


lymereJ:
Thanks! There's some test failures but they were failing before and seem to be NRCAN related. I'll address the failure in the branch directly.

PR #1019, opened 2021-02-19
refactor psz systems

mdahlhausen:
refactor psz-ac and psz-hp systems to use the unitary system object for all psz systems. before, constant volume psz-ac systems were modeled on the air loop.


mdahlhausen:
@yunyangye and @lymereJ I redid the code to accomplish #1009 but a little bit cleaner implementation.  I made a new branch to avoid a messy git log.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2551,110 +2551,53 @@ def model_add_psz_ac(model,
...
-          fan.setAvailabilitySchedule(hvac_op_sch)
-          supplemental_htg_coil.addToNode(air_loop.supplyInletNode) unless supplemental_htg_coil.nil?
-          clg_coil.addToNode(air_loop.supplyInletNode) unless clg_coil.nil?
-          htg_coil.addToNode(air_loop.supplyInletNode) unless htg_coil.nil?
-          fan.addToNode(air_loop.supplyInletNode) unless fan.nil?
-        else
-          OpenStudio.logFree(OpenStudio::Error, 'openstudio.model.Model', 'Invalid fan location')
-          return false
-        end
+        OpenStudio.logFree(OpenStudio::Error, 'openstudio.model.Model', 'Invalid fan_type')
```
lymereJ, line 2565: #1009 was catching invalid fan locations (same as the original code), why do we switch to catching invalid fan type? Shouldn't we catch both?

yunyangye (approved):
I do not see any major difference by comparing my version and it is clear. I approve the change.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
mdahlhausen: we could.  fan_location is only an argument to this method, and it is hardcoded other places in the code that call it.

mdahlhausen: I added a check for fan_location.

mdahlhausen:
@lymereJ @yunyangye can you regenerate the regression / performance tests?

lymereJ (approved):

PR #1022, opened 2021-02-24
Nrcan 279

srgilani:


khaddad:
1/The way the measure is set up now is to pass the name of the chiller to method "modify_chiller_efficiency". I suggest the assignment of the chiller to use be done within the method itself based on the required capacity of the chiller. File "ECMS/hvac_systems.rb" line 1628 has an example of how to pick the right entry in the json file based on capacity

2/Make sure the capacities of the chillers in the json file cover the range of capacities needed for the NECB archetypes.

srgilani:
1) Revised accordingly. See: 
 lib/openstudio-standards/standards/necb/ECMS/hvac_systems.rb
 test/necb/unit_tests/tests/test_ecm_chiller.rb

2) Added other options of chillers. See:
 lib/openstudio-standards/standards/necb/ECMS/data/chiller_set.json
 lib/openstudio-standards/standards/necb/ECMS/data/curves.json

PR #1025, opened 2021-02-27
Dwelling units lighting efficacy requirements

lymereJ:
Section 9.4.4 in 90.1-2016/9 sets a minimum efficacy requirements for hardwired lighting in dwelling units. The proposed changes reduce the LPD in these spaces to capture this new requirement. Additional information in Section B1.3.5 in the [Energy Savings Analysis: ANSI/ASHRAE/IES Standard 90.1-2016 report](https://www.energycodes.gov/energy-savings-analysis-ansiashraeies-standard-901-2016).


mdahlhausen (approved):
This is fine, but just a note - these LPDs are much higher than is typical of residential units.  0.15-0.2 is more common.

PR #1029, opened 2021-03-05
CHW and HW pump control type

lymereJ:
Pump control type for variable speed pumps in the 90.1 prototype is currently assumed to be `Constant Flow` for HW pumps and `VSD DP Reset` for CHW secondary pumps. Pump control requirements actually vary with code versions, pump size, pump head, plant loop type, and climate zone. The proposed changes add pump control requirement for 90.1-2016/9 but also for previous versions of 90.1. Additional information regarding the 90.1-2016/9 implementation are provided in Section B1.2.8 in the [Energy Savings Analysis: ANSI/ASHRAE/IES Standard 90.1-2016 report](https://www.energycodes.gov/energy-savings-analysis-ansiashraeies-standard-901-2016).


leijerry888 (approved):
Code changes look great. Connie summary results look legit.

mdahlhausen (changes requested):
we may want to add the climate zone lookup in lib/openstudio-standards/prototypes/ashrae_90_1/ashrae_90_1_2016/ashrae_90_1_2016.PumpVariableSpeed.rb
to the standards spreadsheet somehow, similar to how economizer limits are handled
We don't have to do this now, and it may make sense to wait as this is just for code readability and there are more pressing issues.

In Prototype.PumpVariableSpeed.rb - the pump_variable_speed_get_control_type method is defaulted to return false.
In other prototype methods, the default is instead the most common option, or something like 90.1-2004 or 90.1-2004.  The intent of the method is to apply a control type regardless of the input model - so the control type should be a valid option, not "false".  I'd be inclined to default whatever is the benchmark in Appendix G 2010.

lymereJ:
> we may want to add the climate zone lookup in lib/openstudio-standards/prototypes/ashrae_90_1/ashrae_90_1_2016/ashrae_90_1_2016.PumpVariableSpeed.rb
> to the standards spreadsheet somehow, similar to how economizer limits are handled
> We don't have to do this now, and it may make sense to wait as this is just for code readability and there are more pressing issues.

This makes sense, I actually debated about doing that but decided to go with including it in the code since it's a small table. I will work on including it in the spreadsheet for future releases.

> In Prototype.PumpVariableSpeed.rb - the pump_variable_speed_get_control_type method is defaulted to return false.
> In other prototype methods, the default is instead the most common option, or something like 90.1-2004 or 90.1-2004. The intent of the method is to apply a control type regardless of the input model - so the control type should be a valid option, not "false". I'd be inclined to default whatever is the benchmark in Appendix G 2010.

Sounds good, I'll make the changes.

lymereJ:
Default behavior of `pump_variable_speed_get_control_type` has been revised.

mdahlhausen (approved):

lymereJ:
@mdahlhausen - Changing the default behavior of `pump_variable_speed_get_control_type` lead to NECB regression test failures. What should we do?

mdahlhausen:
@lymereJ I just overloaded the method for the NECB standard to return false.

@phylroy and @ckirney these are changes to the default pump control type and associated performance curves for the DOE prototypes.  I overloaded the method in NECB standards to no adopt this behavior, but if you want it, you can remove that overload and update your regression models in a future pull request.

PR #1033, opened 2021-03-09
Deactivate occupancy sensor modeling in the code

lymereJ:
Currently, occupancy sensor modeling is done through the usage of different schedules that already account for savings from occupancy sensors. The `model_add_occupancy_sensors` method applies modifiers on top of that, double counting savings from occupancy sensors. The proposed changes "deactivate" this method for the 90.1 prototypes and adds a TODO reference for future improvement.

This is not a 90.1-2016/9 specific changes but since it needs to apply to 90.1-2016/9 the base branch for this PR is the 90.1-2016/9 development branch.


mdahlhausen (approved):
This is fine, but please make an issue addressing the TODO

PR #1038, opened 2021-03-15
Door infiltration calculations

lymereJ:
The proposed changes add methods to handle door infiltration separately from other building infiltration sources. Doing so allow the calculation to be done explicitly in the code as opposed to values in the data JSON files. Requirements for 90.1-2016/9 are also captured. Additional information regarding the 90.1-2016/9 implementation are provided in Section B1.1.3 in the [Energy Savings Analysis: ANSI/ASHRAE/IES Standard 90.1-2016 report](https://www.energycodes.gov/energy-savings-analysis-ansiashraeies-standard-901-2016).


lib/openstudio-standards/prototypes/common/objects/Prototype.Model.rb
```diff
@@ -2159,4 +2158,95 @@ def apply_economizers(climate_zone, model)
...
+  def model_door_infil_flow_rate_open_cfm_ft2(climate_zone)
+    return 9.7875
+  end
+
+  # Add door infiltration
+  #
+  # @param [OpenStudio:Model::Model] OpenStudio model object
+  # @climate_zone [String] Climate zone
+  # @return [Boolean] Returns true if successful, false otherwise or not applicable
+  def model_add_door_infiltration(model, climate_zone)
```
lymereJ, line 2208: There already exist a `add_door_infiltration` method. The main difference is that the one proposed here aims to handle the infiltration calculations for all prototypes. The current one is different for each prototype and uses hard coded values. While this one is targeting 90.1-2016/9 requirements specifically, it could be used in place the the older one, which could then in turn be depreciated.

lymereJ: This triggered the regression test to fail. I don't think we need to update the test results since we know exactly what the discrepancy is.

mdahlhausen (changes requested):
We need to resolve the issue that flow rates standards are set at 75Pa, while openstudio wants the value to be at natural pressure. Merging as is will mean vastly overestimating infiltration.

lymereJ:
> We need to resolve the issue that flow rates standards are set at 75Pa, while openstudio wants the value to be at natural pressure. Merging as is will mean vastly overestimating infiltration.

Ok, I'll uncomment line 2235 in `Prototype.Model.rb`.

mdahlhausen (approved):

PR #1043, opened 2021-03-17
Lighting shutoff

leijerry888:
Add scheduled nighttime light shutoff for prototypes without continuous operations (all prototypes except for apartments, hotels and hospital/outpatient). A set of business hour schedules are developed based on occupancy schedules in use and are updated in both json files and the Google spreadsheet. Additional information regarding the 90.1-2016/9 implementation is provided in Section B1.3.1 in the [Energy Savings Analysis: ANSI/ASHRAE/IES Standard 90.1-2016 report](https://www.energycodes.gov/sites/default/files/documents/02222018_Standard_90.1-2016_Determination_TSD.pdf).


mdahlhausen (commented):
Can you clarify why EMS controls are necessary to implement this change?  And why adding a separate business schedule is necessary?

leijerry888:
> Can you clarify why EMS controls are necessary to implement this change? And why adding a separate business schedule is necessary?

The intention is to have a building-level lighting threshold of 0.02 W/sf during non-business hours. EMS is used to modify lighting power to achieve this intention. Modifying lighting schedules will be difficult with multiple zones sharing the same lighting schedule but having different lpds.
 
The lighting shutoff is based on schedule, according to 90.1 2016 9.4.1.1(i). With this, and the intention of creating building level lights off, having a set of binary valued building level business schedules instead of using zone level occ / lighting schedule is cleaner and we are also thinking of using the business schedules in the future for the implementation of other features.

Details of the modeling strategy is at B.1.3.1 Addendum ah: Egress Lighting Control of [Energy Savings Analysis: ANSI/ASHRAE/IES Standard 90.1-2016 report](https://www.energycodes.gov/sites/default/files/documents/02222018_Standard_90.1-2016_Determination_TSD.pdf), although the OSSTD implementation is somewhat different from what we have done earlier in PNNL prototypes, it is the preferred way of implementing this addendum at this moment.

lymereJ (approved):

mdahlhausen (approved):

PR #1073, opened 2021-04-07
update model_setup_parametric_schedules with new argument

DavidGoldwasser:
expose hoo_var_method that has always been in gather_inputs_parametric_schedule. This setup to default with hours whitch matches current behaior. I plan to use fractional until logic for hours is more robust to avoid out of oder rules.


DavidGoldwasser:
I plan to use the newly exposed argument here for measure that shifts hours of operations by first making parametric schedules form user model. Using fractional instead of hours should be more stable than hours.
https://github.com/NREL/openstudio-calibration-gem/blob/adjust-hours-of-operation/lib/measures/shift_hours_of_operation/measure.rb#L420

mdahlhausen (approved):

mdahlhausen:
@DavidGoldwasser we can include it in this release

PR #1077, opened 2021-04-15
add 90.1 2016 and 2019 data to comstock

mdahlhausen:
changes also include a new space lighting control sheet which includes lighting control details that are added as fields to all standard years


asparke2 (changes requested):
@mdahlhausen we need to repeat the changes that were made (to lighting and plug load schedules mainly, I think) for ComStock.  @amylebar and @JanghyunJK do all the lighting and plug load schedules you modified have "EndUseData" at the end of them?

amylebar:
@asparke2 Yes, all of the schedules we modified should have "EndUseData" at the end.  @mdahlhausen messaged Janghyun, Chris and I about making sure our edits from the end use data work are reflected in the 2016 and 2019 standards versions.  We will get it taken care of.

asparke2 (approved):

PR #1087, opened 2021-05-03
Revise fan power calculation

lymereJ:
Revise fan power calculation to be based on the system retrieved from additional properties, add fan power calculation for non mechanically cooled systems.


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.AirLoopHVAC.rb
```diff
@@ -5,9 +5,135 @@ class ASHRAE901PRM < Standard
...
+  def air_loop_hvac_allowable_system_brake_horsepower(air_loop_hvac)
+    # Get design supply air flow rate (whether autosized or hard-sized)
+    dsn_air_flow_m3_per_s = 0
+    dsn_air_flow_cfm = 0
+    if air_loop_hvac.autosizedDesignSupplyAirFlowRate.is_initialized
+      dsn_air_flow_m3_per_s = air_loop_hvac.autosizedDesignSupplyAirFlowRate.get
+      dsn_air_flow_cfm = OpenStudio.convert(dsn_air_flow_m3_per_s, 'm^3/s', 'cfm').get
+      OpenStudio.logFree(OpenStudio::Debug, 'openstudio.ashrae_90_1_prm.AirLoopHVAC', "* #{dsn_air_flow_cfm.round} cfm = Autosized Design Supply Air Flow Rate.")
+    else
+      puts air_loop_hvac.is_initialized
```
dmaddoxwhite, line 108: Why is this "puts" statement here?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.ZoneHVACComponent.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+    max_air_flow_rate_cfm = OpenStudio.convert(max_air_flow_rate, 'm^3/s', 'ft^3/min').get
+
+    # Set the impeller efficiency
+    fan_change_impeller_efficiency(fan, fan_baseline_impeller_efficiency(fan))
+
+    # Set the motor efficiency, preserving the impeller efficency.
+    # For zone HVAC fans, a bhp lookup of 0.5bhp is always used because
+    # they are assumed to represent a series of small fans in reality.
+    fan_apply_standard_minimum_motor_efficiency(fan, fan_brake_horsepower(fan))
```
dmaddoxwhite, line 90: It looks like fan_brake_horsepower() calls fan_fanpower(), which depends on fan.pressureRise. Since final pressureRise has not yet been set, has it been initialized to a reasonable value at this point?

dmaddoxwhite (approved):
Looks good! I left a couple of questions.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.AirLoopHVAC.rb
lymereJ: Good catch, not sure what it's doing there. Probably a left over from testing I was doing before having access to Ruby debugger. I removed it in the last commit.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.ZoneHVACComponent.rb
lymereJ: It uses the pressure rise from the `ashrae_90_1_prm.fans.json` file.

lymereJ:
@dmaddoxwhite - I wonder if [this method](https://github.com/NREL/openstudio-standards/blob/99cf713750661fe7d2082739f251269c2dfd9140/lib/openstudio-standards/standards/Standards.Fan.rb#L219L276) actually doesn't captures what's required by Section G.3.1.2.9. The section requires that the motor efficiency be the efficiency from Table G3.9.1 for the next motor size greater than the bhp. Here's an example that I came across: if `motor_hp` is 3.5 hp [this line](https://github.com/NREL/openstudio-standards/blob/99cf713750661fe7d2082739f251269c2dfd9140/lib/openstudio-standards/standards/Standards.Fan.rb#L252) would return `5` because it would fall in [this category](https://github.com/NREL/openstudio-standards/blob/AppendixG_Dev/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/data/ashrae_90_1_prm_2019.motors.json#L44-L51) because its `maximum_capacity` is `4.999`. Then, [this line](https://github.com/NREL/openstudio-standards/blob/99cf713750661fe7d2082739f251269c2dfd9140/lib/openstudio-standards/standards/Standards.Fan.rb#L267) adds `0.01` to the nominal hp putting it in the next category when looking the efficiency value, which now returns the efficiency from [this category](https://github.com/NREL/openstudio-standards/blob/AppendixG_Dev/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/data/ashrae_90_1_prm_2019.motors.json#L54-L61). In short the method seems to be retrieving the efficiency of the next two motor sizes greater than the bhp.

I wonder if [this line](https://github.com/NREL/openstudio-standards/blob/99cf713750661fe7d2082739f251269c2dfd9140/lib/openstudio-standards/standards/Standards.Fan.rb#L252) shouldn't actually be `nominal_hp = motor_bhp.to_f.round(1)` or if it actually even needs to be there. Making that change would probably impact the prototypes and the implementation of the PRM for the older versions of 90.1.

What do you think?

dmaddoxwhite:
@lymereJ - I think you are right to modify the line.  Would it make sense to remove [this line](https://github.com/NREL/openstudio-standards/blob/99cf713750661fe7d2082739f251269c2dfd9140/lib/openstudio-standards/standards/Standards.Fan.rb#L267), which adds the 0.01 to nominal_hp? The model_find_object function is already doing some adjustements for when the actual value is near the boundary of the min/max of the data set.
In model_find_objects, first [here](https://github.com/NREL/openstudio-standards/blob/99cf713750661fe7d2082739f251269c2dfd9140/lib/openstudio-standards/standards/Standards.Model.rb#L1655) and again [here](https://github.com/NREL/openstudio-standards/blob/99cf713750661fe7d2082739f251269c2dfd9140/lib/openstudio-standards/standards/Standards.Model.rb#L1662).  In fact, I think the above corrections create a problem for a case where the boundary is not an integer. For example, if the target is 1.51, it rounds to 2.0, and the adjustment turns that into 2.02, which selects from the 2.0 to 2.999 range, which becomes 3.0 nominal capacity. Maybe the rounding needs to happen to the nearest tenth instead of to a whole number.

lymereJ:
@dmaddoxwhite - I agree, I made some modifications along those lines. Now, if the target is 1.51, the code will select from the 1.5 to 1.999 and return 2 as the nominal hp (i.e. the next size up). Please, let me know what you think.

PR #1108, opened 2021-06-12
Update performance regression test results using OS 3.2.0

lymereJ:
Update performance regression test results using OS 3.2.0. CI uses 3.2.1-rc1.


mdahlhausen:
we'll re-update the regression tests with a new branch

mdahlhausen:
I'll leave this branch up until we've merged a new set of regression tests

PR #1112, opened 2021-06-21
ASHRAE 62.1 Simplified Procedure

lymereJ:
The proposed changes update the method to calculate the minimum damper position and system design outdoor air based on the ASHRAE 62.1 Simplified Procedure (as opposed to the method in Appendix A) as required by ASHRAE 90.1-2019. Additional information regarding the implementation are provided in Section B2.7 in the [Energy Savings Analysis: ANSI/ASHRAE/IES Standard 90.1-2019 report](https://www.energycodes.gov/sites/default/files/documents/20210407_Standard_90.1-2019_Determination_TSD.pdf).

The proposed changes also improve the existing method by letting EnergyPlus size the system outdoor air intake based on a minimum zone ventilation of 0.6 (not possible before OS 3.0, follow up on the issue identified in #776), and remove redundant code.

Note: It was noticed that the maximum air terminal air flow calculated from the sizing run could be different from the annual simulation run. The reason for that is that sizing runs are currently done without daylighting control, as they should, but autosizing calculation for the annual run are done with daylighting control because OS currently doesn't allow setting an availability schedule to daylighting control, see [here](https://github.com/NREL/OpenStudio/issues/4364).

Note 2: This PR is affected by https://github.com/NREL/OpenStudio/issues/4436.


lib/openstudio-standards/standards/ashrae_90_1/ashrae_90_1_2010/ashrae_90_1_2010.AirLoopHVAC.rb
```diff
@@ -1,27 +1,6 @@
...
-    # exclusion: for Outpatient: (1) both AHU1 and AHU2 in 'DOE Ref Pre-1980' and 'DOE Ref 1980-2004'
-    # (2) AHU1 in 2004-2019
-    # TODO refactor: move building-type-specific code to Prototype classes
-    if air_loop_hvac_multizone_vav_system?(air_loop_hvac) && !(air_loop_hvac.name.to_s.include? 'Outpatient F1')
-      air_loop_hvac_adjust_minimum_vav_damper_positions(air_loop_hvac)
-    end
-
-    return true
-  end
-
```
lymereJ, line 24: Remove redundant code.

lib/openstudio-standards/standards/ashrae_90_1/ashrae_90_1_2019/ashrae_90_1_2019.AirLoopHVAC.rb
```diff
@@ -453,4 +432,173 @@ def air_loop_hvac_energy_recovery_ventilator_flow_limit(air_loop_hvac, climate_z
...
+    max_oa_ems_prog_manager.addProgram(max_oa_ems_prog)
+
+    # Hard-size the sizing:system
+    # object with the calculated min OA flow rate
+    sizing_system = air_loop_hvac.sizingSystem
+    sizing_system.setDesignOutdoorAirFlowRate(v_ot)
+    sizing_system.setSystemOutdoorAirMethod('ZoneSum')
+
+    return true
+  end
```
lymereJ, line 603: New method that sets the MDP based on `v_oz` and `v_ot` based on a simplified `e_v` calculation.

lib/openstudio-standards/prototypes/common/objects/Prototype.SizingSystem.rb
```diff
@@ -39,15 +39,23 @@ def adjust_sizing_system(air_loop_hvac,
...
-    # sizing_system.setSystemOutdoorAirMethod("VentilationRateProcedure")
+    sizing_system.setSystemOutdoorAirMethod('VentilationRateProcedure')
     # Set the minimum zone ventilation efficiency to be 0.6
     air_loop_hvac.thermalZones.sort.each do |zone|
       sizing_zone = zone.sizingZone
-      # It is not yet possible to adjust the minimum zone ventilation efficiency
-      # @todo, update this section when OS allows to adjust minimum zone ventilation efficiency
-      # In EnergyPlus this is done through the DesignSpecification:ZoneAirDistribution object
-      # which is then assigned to a Sizing:Zone object
+      sizing_zone.setDesignMinimumZoneVentilationEfficiency(0.6)
```
lymereJ, line 58: Sets the minimum `e_vz` which wasn't possible prior to OS 3.0.

mdahlhausen (changes requested):
- Add log/warning message for method that isn't supported for OS versions < 3.0 (add version check in code).
- check alignment between system outdoor air method in sizing system and controller mechanical ventilation

lymereJ:
> * Add log/warning message for method that isn't supported for OS versions < 3.0 (add version check in code).
> * check alignment between system outdoor air method in sizing system and controller mechanical ventilation

I added a warning for the unsupported method.

As I expected, the fact that the controller mechanical ventilation is created by default and cannot be unassigned from the controller outdoor air (unless I'm mistaken) can potentially be an issue because if the OA flow rate specified at the sizing system level is lower than what is recalculated by the controller mechanical ventilation using the "zone sum" approach, the later is used during the simulation. Since this is can be an issue now, and will be if this PR is merged, I would be inclined to merge it as is, file an issue on the OpenStudio Github page, and reference this PR. What do you think?

mdahlhausen (approved):

PR #1122, opened 2021-07-02
Adding Iqaluit weather files. 

IainMacdonald:
Added weather files for Iqaluit to the data/weather folder. Also updated the btap/WeatheData files - not sure if these are required or not.

Assumed that deltadb oin the json file is the difference between Table C1 altitude and E+ weather file altitude (with +ive being table C1 higher).

I've assigned Chris to review - hope this is OK.


IainMacdonald:
@ckirney you should be good to go with the merge now. I added the expected results for the weather test on the new file.

Issue #1013, opened 2021-02-03
Add rubocop check to tests


mdahlhausen:
Add rubocop to the tests for pull requests / commits as requested by @lymereJ

mdahlhausen:
This one is tricky - there are some warning messages that show up simply because the methods are long.  I think this is something we should check with code commits, but I don't think we should require it as part of the tests.

PR #1123, opened 2021-07-02
Comprehensive documentation of standards methods

mdahlhausen:
While this seems like a large pull request because it touches almost all of the code, it only adds documentation to existing methods and therefore doesn't change regressions.  This resolves many issues of undocumented methods and rubocop errors from running the rubocop and doc rake tasks respectively. I may go in later and add some more detailed examples to key, outward facing standards methods.

@ckirney @phylroy @lymereJ @asparke2 tagging you so you are aware of the documentation changes.

I left the NRCAN/NECB sections of code unchanged (with one exception in removing one unncessary method argument in one method).
I recommend that the NRCAN team run the doc and rubocop rake tasks and resolve the rubocop errors and document undocumented methods.  There are also a lot of NECB specific orphaned methods in lib/openstudio-standards/utilities that can be deleted.


lymereJ (approved):
That's great! Thanks for doing that.  It would be great if there was a way to flag improper/insufficient formatting for PRs so we don't to do that again. I guess this is linked to #1013.

PR #1131, opened 2021-07-09
Gem update

mdahlhausen:
Updates the gemspec to remove the google-api-client gem and its dependencies.  Uses the newer google_drive gem instead.

Also update rubyXL and nokogiri, which @ckirney raised has having security issues with the older versions.

Luckily, the google_drive gem is much simpler than the older version, so this was an easy refactor that got rid of a decent chunk of code.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -175,20 +175,23 @@ namespace :data do
   spreadsheet_titles = spreadsheets_ashrae + spreadsheets_deer + spreadsheets_comstock + spreadsheets_cbes
   spreadsheet_titles = spreadsheet_titles.uniq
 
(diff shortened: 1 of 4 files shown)


ckirney:
I pulled the changes into the the nrcan_master_merge branch and checked it with Trivy.  Everything looks good.

ckirney (commented):
As mentioned in my other post.  I pulled the changes into the nrcan_master_merge branch, updated the gems and checked with Trivy.  The new gems look are good and there are no security issues.  I did not check anything beyond the revised gemspec.

asparke2 (approved):

PR #1155, opened 2021-07-27
remove redundant dependency calls

mdahlhausen:
remove direct require calls to developer dependencies in btap_datapoint.rb


mdahlhausen:
@phylroy to review and potential add back in dependencies to the initialize method

mdahlhausen:
@phylroy I merged this pull request into the main branch to keep that branch functional.
You can continue working on this branch if you want to re-enable the require statements if needed.
Note that the SSL constant setting cannot be done in a method, as that makes it a dynamic assignment (could change each time method is called), which is not allowed in the ruby interpreter.

ckirney:
I want to clarify your statement about continuing to work on this branch if we want to re-enable require statements.  Is that, we can re-enable them in the code on master, or just our own branch?

mdahlhausen:
you can add the requires back in, but remove the btap_datapoint.rb dependency in lib/openstudio_standards.rb

PR #1178, opened 2021-08-27
Merge nrcan branch changes into master

ckirney:


mdahlhausen (changes requested):
I've reviewed the merge.  See comments below.  You may want to run rubocop once more to confirm you have proper syntax throughout.
There were two other comments we may want to address before the merge - the need to 'touch' the weather file in simulation.rb, and the require calls and hardcoded paths in btap_cli/test/space_type.rb.  I'm mark this as request changes for now and will approve/merge once we've got those two comments addressed.

btap/envelope.rb lines 837, 839 - rubocop or vs ||

lib/openstudio-standards/standards/necb/NECB2011/qaqc/necb_qaqc.rb line 630 - rubocop or vs ||

standards/necb/ECMS/hvac_systems.rb line 2889 - you may want to include defaults here for the optional arguments, as there is no error checking if left blank

lib/openstudio-standards/utilities/simulation.rb line 104 - this comment is quite concerning.  Is this something we should address before merging?

utilities/btap_cli/tests/space_type.rb - there are two require calls here.  If this test is in standards, both of these gems should already be loaded so these require calls are redundant.  I also think the hardcoded os_standards_local_lib_path on line 6 and others on line 14 and 60 will cause you some issues.  Use "#{File.dirname(__FILE__) or something like that instead.

IainMacdonald:
@mdahlhausen in relation to the 'touch' hack some more details. I can describe this in more detail next week.
We generally run our models via PAT. We have a custom version of server running (essentially it ties into nrcan branch of standards so is currently linked to 3.0.1 of standards). At some point the simulations on server stopped working. With a whole load of debugging I discovered that the model was not finding the weather file (despite it being present in the folder and the correct path being in the model). When I ran the osw in the worker node using the CLI everything worked (using the CLI command from the next line of code). I cannot remember all the things I did (copying the weather file etc) but eventually stumbled on just 'touching' the file to get it to work. As we know this does nothing other than updating the date stamp on the file so I really cannot explain why it works, nor can I explain why it was not working before! 
Let me now if there are any specific things yo want me to check before the meeting. I can send our scripts through (they are all probably based on NRCan/NRELs if you go back far enough - but we have a lot of changes) if that would help - but its probably easier for me to talk through what we've done.

mdahlhausen:
@tijcolem see the above about a potential issue with OS server and the OS CLI

ckirney:
I updated the above issue to resolve some of the issues identified by @mdahlhausen including:
-The identified rubocop issues were resolved
-The standards/necb/ECMS/hvac_systems.rb line 2889 issue now covers all possible options
-The utilities/btap_cli/tests/space_type.rb file was moved to lib/openstudio-standards/btap/utilities/ since it is a script developed to prepare a specific set of files to work better with BTAP so some of the hard coded paths cannot necessarily be easily generalized.
-The touch statement was removed from lib/openstudio-standards/utilities/simulation.rb line 104 but the issue it intended to resolve will be discussed further and hopefully a better solution can be developed

mdahlhausen (approved):
Thanks for making the requested changes.  I've tagged @tijcolem on this thread to continue the touch file issue.  We will address that in a separate branch/pull-request.

tijcolem:
@IainMacdonald To me, the only explanation where usage of touch would make sense to me is if the file just didn't exist in the run folder and some check was preventing it to proceed w/o the presence of the file. touch will create an empty file if it doesn't exit.  

If you have a the code handy, maybe you can do a long file listing in the run directory before you call touch and see associated permissions and even cat out the contents of the weather file.  

Also, you can open up an issue at https://github.com/NREL/OpenStudio-server with steps to reproduce and we can look into it and figure out what is causing it.

IainMacdonald:
Hi @tijcolem - I agree. The file existed and had the correct contents - from memory I had tried doing another copy etc which worked but made no sense (not that touching the file makes much sense either). Its easy enough for me to reproduce with my tool chain so can demo that. I'll also try and document it with the main branch of server (but I may be short of time to do that).

PR #1199, opened 2021-09-08
Updating require statement in btap_datapoint.rb

ckirney:
Added back a git-revision require statement in btap_datapoint.  The git-revision gem is already part of the gemspec in master so this should not cause any issues with other users of openstudio-standards and will resolve some NRCan issues.


mdahlhausen (changes requested):
If this require is necessary, [remove btap_datapoint.rb from openstudio_standards.rb](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards.rb#L74).

The gemspec is for *development* dependencies.  You must run bundle install to establish a development environment to run many scripts in openstudio-standards.  For example,  [manage_OpenStudio_Standards.rb](https://github.com/NREL/openstudio-standards/blob/master/data/standards/manage_OpenStudio_Standards.rb) requires the rubyXL and google-drive gems, and these are listed as development dependencies in the gemspec.  But manage_OpenStudio_Standards.rb is not required in [openstudio_standards.rb](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards.rb) and would not run in the OpenStudio CLI without first running bundle install on the openstudio-standards gemspec.  The core of openstudio-standards (what is required in openstudio_standards.rb) must rely on solely the OpenStudio *production* dependencies which is the OpenStudio CLI.  You can see what gems are available there by running `openstudio gem_list` in terminal.

ckirney:
Okay.  I will close this pull request then and remove it from future ones.

PR #1210, opened 2021-09-14
Add 90.1-2019 prototypes

lymereJ:
The proposed changes add the ASHRAE Standard 90.1-2019 prototypes to openstudio-standards. This includes all pull requests with the `90.1-2016/9` tag that have been submitted previously.


lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -6475,4 +6475,227 @@ def model_add_hvac_system(model,
     rename_air_loop_nodes(model)
     rename_plant_loop_nodes(model)
   end
+
+  # This method will add an swh water fixture to the model for the space.
+  # if the it will return a water fixture object, or NIL if there is no water load at all.
+  def model_add_swh_end_uses_by_spaceonly(model, space, swh_loop)
```
mdahlhausen, line 6481: this should go in the service water heating section

lib/openstudio-standards/prototypes/common/objects/Prototype.Model.rb
```diff
@@ -1239,6 +1250,195 @@ def model_add_guestroom_vacancy_controls(model, building_type)
...
+    return true unless (template == '90.1-2016') || (template == '90.1-2019')
+
+    # Guestrooms are currently only included in the small and large hotel prototypes
+    return true unless (building_type == 'LargeHotel') || (building_type == 'SmallHotel')
+
+    OpenStudio.logFree(OpenStudio::Info, 'openstudio.model.Model', 'Started Adding Guestroom Ventilation Schedules')
+
+    # Define guestroom occupied maps
+    # List of all spaces that represent vacant rooms
+    guestroom_occupied_map = {
```
mdahlhausen, line 1270: this section should be in the custom prototype section for each building type

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
mdahlhausen: I deleted this section; it is overloaded in the NECB code and only used there, so it doesn't need to be in hvac_systems.rb

mdahlhausen (approved):

lib/openstudio-standards/prototypes/common/objects/Prototype.Model.rb
mdahlhausen: This is the prototype section.  While I generally would prefer to see this in the custom adjustments under each building type, it is shared and only applies to the prototype buildings.  It can stay as is.

PR #1213, opened 2021-09-17
clean up of necb 2020 files to make use of inheritance

IainMacdonald:
removed the methods that can be safely inherited from necb2017 and earlier.


IainMacdonald:
@ckirney - I'm not sure about line 129 of building envelope ("when /skylight/") this was a guess as to what the update should be here. From what I can see in our create_prototype measure the correct data is being applied for skylights.

PR #1229, opened 2021-11-15
Fix autosized minimum OA flow rate for OA controllers and ERV SAT control

lymereJ:
Remove multipliers from autosizedMinimumOutdoorAirFlowRate and set SAT control for ERV to true when ERV are required.


mdahlhausen:
can you update the regression test?

ckirney:
The relevant NRCan regression tests have been updated to reflect the ERV changes.  All NRCan tests should now pass.

PR #1231, opened 2021-11-19
Fix the bug that adding 'PumpedCondenser' type heat pump water heater…

KaiyuSun:
… will generate a redundant unused tank (WaterHeaterMixed).

OpenStudio::Model::WaterHeaterHeatPump.new will automatically generate a new WaterHeaterMixed as the tank and attach to the HPWH. But later in the method another WaterHeaterMixed is created and linked to the HPWH, which leaves the initial one unused and redundant. Same thing for the coil (WaterHeatingAirToWater). Fixed the bug.

Another issue is that the evaporator air flow rate was hardsized, which may cause fatal error when the capacity doesn't fit the hardsized flow rate. It should be autocalculated as it varies with the capacity.

@mdahlhausen Please help review. Thanks!

### DIFF
--- a/lib/openstudio-standards/prototypes/common/objects/Prototype.ServiceWaterHeating.rb
+++ b/lib/openstudio-standards/prototypes/common/objects/Prototype.ServiceWaterHeating.rb
@@ -386,7 +386,6 @@ class Standard
     end
 
     # set heat pump water heater properties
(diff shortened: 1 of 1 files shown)


mdahlhausen (approved):
Thanks for catching this.  I made a minor edit to remove the line that sets the evaporatorairflowrate.

KaiyuSun:
@mdahlhausen Thank you for the quick action!

PR #1238, opened 2021-12-02
Preheat coil

weilixu:
Complete function to add setpointmanager:scheduled to heating coils in the system type 5, 6, 7, and 8.
The changes are documented in the commits.

Complete the unit test functions to test system type 5, 6, 7, and 8 test cases. All passed.

Updated the document in preheat_coil.md


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -590,4 +590,58 @@ def thermal_zone_prm_lab_delta_t(thermal_zone)
   def model_get_fan_power_breakdown
     return true
   end
+
+  def model_set_spm(model, thermalZones, coil)
```
lymereJ, line 594: We need to add YARD comments for this methods. I would also rename it to something more specific, `model_set_spm` is too broad, maybe something like `model_set_central_preheat_coil_spm()`?

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2018,25 +2026,28 @@ def model_add_pvav(model,
...
+      htg_coil = create_coil_heating_water(model, hot_water_loop,
+                                           air_loop_node: air_loop.supplyInletNode,
+                                           name: "#{air_loop.name} Main Htg Coil",
+                                           rated_inlet_water_temperature: hw_temp_c,
+                                           rated_outlet_water_temperature: (hw_temp_c - hw_delta_t_k),
+                                           rated_inlet_air_temperature: dsgn_temps['prehtg_dsgn_sup_air_temp_c'],
+                                           rated_outlet_air_temperature: dsgn_temps['htg_dsgn_sup_air_temp_c'])
     end
 
+    # 2021-11-23
```
lymereJ, line 2047: We don't need this comment. There's multiple instances of it.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -2018,25 +2026,28 @@ def model_add_pvav(model,
...
+    # set the setpointmanager for preheat coil for 2019-PRM stable baseline
```
lymereJ, line 2048: I would not specifically mention the 2019 PRM here but be more general: `#set the setpointmanager for the central/preheat coil if required`. There's multiple instances of it.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -590,4 +590,58 @@ def thermal_zone_prm_lab_delta_t(thermal_zone)
...
+                                                           name = "#{coil_name} Setpoint Temp - #{preheat_setpoint_f.round}F")
+    preheat_coil_manager = OpenStudio::Model::SetpointManagerScheduled.new(model, preheat_coil_sch)
+    preheat_coil_manager.setName("#{coil_name} Preheat Coil Setpoint Manager")
+
+    if coil.to_CoilHeatingWater.is_initialized
+      preheat_coil_manager.addToNode(coil.airOutletModelObject.get.to_Node.get)
+    elsif coil.to_CoilHeatingElectric.is_initialized
+      preheat_coil_manager.addToNode(coil.outletModelObject.get.to_Node.get)
+    elsif coil.to_CoilHeatingGas.is_initialized
+      preheat_coil_manager.addToNode(coil.airOutletModelObject.get.to_Node.get)
```
lymereJ, line 642: I understand that you have this to align with the methods that create systems in `Prototype.Model.rb`. I think we should add a warning message here saying that preheat coils in the baseline should only be hydronic or electric and perhaps add a reference to the PRM-RM. The warning should technically never appear but if for some reason it would, it'll be nice to know that there's an issue with the baseline system creation via the log file.

test/90_1_prm/test_appendix_g_prm.rb
```diff
@@ -1310,6 +1363,64 @@ def check_if_sz_vav(model, zone)
...
+      # Concatenate modifier functions and arguments
+      mod_str = mod.flatten.join('_') unless mod.empty?
+
+      htg_coil_node_list = []
+      model_baseline.getAirLoopHVACs.each do |airloop|
+        # Baseline system type identified based on airloop HVAC name
+        if airloop.name.to_s.include?('Sys5') ||
+            airloop.name.to_s.include?('Sys6') ||
+            airloop.name.to_s.include?('Sys7') ||
+            airloop.name.to_s.include?('Sys8')
```
lymereJ, line 1382: Nothing wrong with that block of code but we had decided to move away from using the name to identify the system type and use additional properties (available at the thermal zone and air loop level) to get it. Here's an example: `system_type = air_loop_hvac.additionalProperties.getFeatureAsString('baseline_system_type').get`

test/90_1_prm/test_appendix_g_prm.rb
```diff
@@ -1310,6 +1363,64 @@ def check_if_sz_vav(model, zone)
...
+          airloop.supplyComponents.each do |equip|
+            if equip.to_CoilHeatingWater.is_initialized
+              htg_coil = equip.to_CoilHeatingWater.get
+              heating_coil_outlet_node = htg_coil.airOutletModelObject.get.to_Node.get
+            elsif equip.to_CoilHeatingElectric.is_initialized
+              htg_coil = equip.to_CoilHeatingElectric.get
+              heating_coil_outlet_node = htg_coil.outletModelObject.get.to_Node.get
+            elsif equip.to_CoilHeatingGas.is_initialized
+              htg_coil = equip.to_CoilHeatingGas.get
+              heating_coil_outlet_node = htg_coil.airOutletModelObject.get.to_Node.get
```
lymereJ, line 1395: Similar comment as above, I think that if the system type is 5 or 7 we should fail the test as it should be hydronic.

test/90_1_prm/test_appendix_g_prm.rb
```diff
@@ -1935,13 +2046,15 @@ def test_create_prototype_baseline_building
...
-      'hvac_baseline',
+#      'hvac_baseline',
 #      'hvac_psz_split_from_mz',
+#      'plant_temp_reset_ctrl',
 #      'sat_ctrl',
 #      'number_of_boilers',
 #      'number_of_chillers',
 #      'number_of_cooling_towers',
 #      'hvac_sizing',
+      'preheat_coil_ctrl'
```
lymereJ, line 2057: Could you please reactive all tests? I don't think your changes should make other tests fail but for future development it'd be nice to have them all active.

lymereJ (changes requested):
Overall, it looks great. I just have a few concerns and suggested modifications that I've added as comments. Thanks for putting the markdown document together, they're great for the review!

lib/openstudio-standards/standards/ashrae_90_1_prm/docs/set_hw_and_chw_supply_water_temp_reset_control.md
```diff
@@ -0,0 +1,24 @@
+Apply water supply temperature reset control to hot water loop and chilled water loop.
```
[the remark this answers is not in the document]
weilixu: I am not sure. It could be I accidentally added this md to the preheat coil branch when I was working on both branches.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -590,4 +590,58 @@ def thermal_zone_prm_lab_delta_t(thermal_zone)
...
+          elsif setpoint_sch.to_ScheduleCompact.is_initialized
+            setpoint_sch = setpoint_sch.to_ScheduleCompact.get
+            setpoint_c = schedule_compact_annual_min_max_value(setpoint_sch)['max']
+          end
+          if setpoint_c > max_heat_setpoint
+            max_heat_setpoint = setpoint_c
+          end
+        end
+      end
+    end
```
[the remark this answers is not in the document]
weilixu: The decision for such implementation was based on:
1. An assumption that winter design day setpoint shall be the max value of an annual heating setpoint schedule.
2. Existing functions available in the code base for search for finding the annual maximum heating setpoint.

It is may not straightforward but should work well with the above assumptions.

@dmaddoxwhite could you take a look at this comment?

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
weilixu: Remove the comments

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
weilixu: Updated

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
weilixu: Added OpenStudio::Warn

test/90_1_prm/test_appendix_g_prm.rb
weilixu: Add this as a failure case - for system 5-8

test/90_1_prm/test_appendix_g_prm.rb
weilixu: Activated all test cases.

test/90_1_prm/test_appendix_g_prm.rb
weilixu: Updated

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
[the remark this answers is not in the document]
weilixu: @lymereJ @dmaddoxwhite 
The new code is pushed to the repo for merge. The setpoint search is relocated to the Standard.Model.rb as a utility function.
Getting min max values from design day is implemented for both ruleset and compact.
Constant schedule is ignored in this update because the schedule has only one value.

weilixu:
@lymereJ @dmaddoxwhite  Updated the design day min max search - ready for review and let me know if this PR is good for merge. Thanks

lib/openstudio-standards/standards/Standards.ScheduleCompact.rb
```diff
@@ -1,6 +1,50 @@
...
+      end
+
+      str = eg.getString(0)
+      if str.is_initialized
+        prev_str = str.get.downcase
+        if prev_str.include?('for:')
+          # Process a new day schedule, turn the flag off.
+          design_day_flag = false
+          # in the same line, if there is design day label and matches the type, turn the flag back on.
+          if (prev_str.include?('winterdesignday') && type == 'heating') || (prev_str.include?('summerdesignday') && type == 'cooling') || prev_str.include?('alldays')
```
lymereJ, line 29: Small suggestion: instead of using `heating` and `cooling` as `type`, perhaps we could use `winter` and `summer` so then that line could be simplified to:
```
if (prev_str.include?(type) || prev_str.include?('alldays')
```

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -7350,4 +7350,40 @@ def model_get_fan_power_breakdown
...
+        # for constant schedule, there is only one value, so the annual should be equal to design condition.
+        setpoint_min_max = schedule_constant_annual_min_max_value(schedule)
+      elsif schedule.to_ScheduleCompact.is_initialized
+        schedule = schedule.to_ScheduleCompact.get
+        setpoint_min_max = schedule_compact_design_day_min_max_value(schedule, type)
+      end
+      return setpoint_min_max
+    end
+
+    return { 'min' => 999.9, 'max' => 999.9 }
```
lymereJ, line 7387: Should we trigger an error if this occurs? Or at least a log. It's unlikely that it'll happen but if it does we would be assuming a very high values.

lib/openstudio-standards/standards/Standards.ScheduleCompact.rb
```diff
@@ -1,6 +1,50 @@
...
+            design_day_flag = true
+          end
+        end
+      end
+    end
+
+    # Error if no values were found
+    if vals.size.zero?
+      OpenStudio.logFree(OpenStudio::Error, 'openstudio.standards.ScheduleCompact', "Could not find any value in #{schedule_compact.name} design day schedule when determining min and max.")
```
lymereJ, line 38: Looks like you added a log here 👍 (see first comment)

lymereJ (approved):
Looks good as is! I made a few comments, feel free to address them if you agree.

lib/openstudio-standards/standards/Standards.ScheduleCompact.rb
weilixu: Addressed, good suggestion!

lib/openstudio-standards/standards/Standards.Model.rb
weilixu: Added. It is unlikely to reach this line, so if it happens, an Error would be appropriate. I added the Error log before return value

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -590,4 +590,54 @@ def thermal_zone_prm_lab_delta_t(thermal_zone)
...
+
+    max_heat_setpoint_f = OpenStudio.convert(max_heat_setpoint, 'C', 'F').get
+    preheat_setpoint_f = max_heat_setpoint_f - 20
+    preheat_setpoint_c = OpenStudio.convert(preheat_setpoint_f, 'F', 'C').get
+    # in this situation, we hard set the temperature to be 52 F or 11.1 C
+    # (ASHRAE 90.1 Room heating stepoint temperature is 72 F)
+    preheat_setpoint_c = 11.1 if preheat_setpoint_c == 0.0
```
dmaddoxwhite, line 623: What is this line doing? Why would preheat_setpoint_c be zero?

dmaddoxwhite (commented):
Looks good! I added a question regarding the new spm routine.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
weilixu: Ah, thanks for catching this line.

The line 603 sets a default value to max_heat_setpoint = 0.0,
so the line 623 originally was coded to prevent failure from searching the max design setpoint. But I forgot to update the condition after adding the unit conversions.

In this case, I can move this code to line [617](https://github.com/NREL/openstudio-standards/blob/383a1e050ecdfbb350815b850bb8e51ef16d8f37/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb#L617):
max_heat_setpoint= 72 if max_heat_setpoint== 0.0 (assuming no heating setpoint temperature is lower than 0 degree C.
and the rest will be the same.

dmaddoxwhite: Sounds good!

dmaddoxwhite (approved):

PR #1242, opened 2021-12-10
Reopen the issue. OpenStudio automatically hardsize the evaporator ai…

KaiyuSun:
…r flow rate of the DX water heating coil and HPWH while creating the objects. Simply removing the hardsizing at the OS Standards gem side can't solve the problem, have to specifically assign "autocalculate" and "autosize" to make them vary with capacity.

@mdahlhausen


mdahlhausen:
@lymereJ will this impact the prototypes?

lymereJ:
I don't think so, it seems that this for heat pump water heaters, I don't think that prototypes currently use heat pump water heaters.

PR #1247, opened 2021-12-23
Osstd 261 cz3 baseline window construction bug

weilixu:
Fix the CZ3 Window construction bug - run CI for a complete set test

### DIFF
--- a/lib/openstudio-standards/standards/Standards.Model.rb
+++ b/lib/openstudio-standards/standards/Standards.Model.rb
@@ -3921,6 +3921,28 @@ class Standard
                                 'building_category' => building_category,
                                 'minimum_percent_of_surface' => minimum_percent_of_surface,
                                 'maximum_percent_of_surface' => maximum_percent_of_surface)
+
+      # 12/14/2021 added to test data pull from sub climate zones for transparent surfaces
+      if !props
+        if climate_zone_set == "ClimateZone 3"
+          # for PRM, climate zone 3 should test on 3A, 3B and 3C for transparent surfaces
+          search_test_set = ["A", "B", "C"]
+          for test in search_test_set
+            props = model_find_object(standards_data['construction_properties'],
+                                      'template' => template,
+                                      'climate_zone_set' => climate_zone_set+test,
+                                      'intended_surface_type' => intended_surface_type,
+                                      'standards_construction_type' => standards_construction_type,
+                                      'building_category' => building_category,
+                                      'minimum_percent_of_surface' => minimum_percent_of_surface,
+                                      'maximum_percent_of_surface' => maximum_percent_of_surface)
+            if props
+              # Found the transparent surfaces, stop the searching.
+              break
+            end
+          end
+        end
+      end
     else
       props = model_find_object(standards_data['construction_properties'],
                                 'template' => template,


lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -3921,6 +3921,28 @@ def model_find_and_add_construction(model, climate_zone_set, intended_surface_ty
                                 'building_category' => building_category,
                                 'minimum_percent_of_surface' => minimum_percent_of_surface,
                                 'maximum_percent_of_surface' => maximum_percent_of_surface)
+
+      # 12/14/2021 added to test data pull from sub climate zones for transparent surfaces
```
lymereJ, line 3925: I think that this comment could be removed or at least rephrased to make the intent of the code below a bit more clear. The date can be removed.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -3921,6 +3921,28 @@ def model_find_and_add_construction(model, climate_zone_set, intended_surface_ty
...
+      if !props
+        if climate_zone_set == "ClimateZone 3"
```
lymereJ, line 3927: Looks like there's not `else` so these could probably be on the same line.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -3921,6 +3921,28 @@ def model_find_and_add_construction(model, climate_zone_set, intended_surface_ty
...
+                                      'climate_zone_set' => climate_zone_set+test,
+                                      'intended_surface_type' => intended_surface_type,
+                                      'standards_construction_type' => standards_construction_type,
+                                      'building_category' => building_category,
+                                      'minimum_percent_of_surface' => minimum_percent_of_surface,
+                                      'maximum_percent_of_surface' => maximum_percent_of_surface)
+            if props
+              # Found the transparent surfaces, stop the searching.
+              break
+            end
```
lymereJ, line 3942: Unless I'm missing something I think that there is a potential issue with the proposed changes. For example, if the targeted climate zone is 3B and a specific code version (or `template`) has different requirements for 3A and 3B then it seems that the proposed changes would return the requirement for climate zone 3A and not 3B. We might not currently have different requirements for 3A and 3B for the PRM but because the proposed changes are in `Standards.Model` which is used by other projects within openstudio-standards it might be an issue at some point.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -3921,6 +3921,28 @@ def model_find_and_add_construction(model, climate_zone_set, intended_surface_ty
                                 'building_category' => building_category,
                                 'minimum_percent_of_surface' => minimum_percent_of_surface,
                                 'maximum_percent_of_surface' => maximum_percent_of_surface)
+
+      # 12/14/2021 added to test data pull from sub climate zones for transparent surfaces
+      if !props
+        if climate_zone_set == "ClimateZone 3"
+          # for PRM, climate zone 3 should test on 3A, 3B and 3C for transparent surfaces
```
lymereJ, line 3928: This comment should probably be pushed in the first one but without mentioning PRM (it might already be the case for other climate zone sets).

lib/openstudio-standards/standards/Standards.Model.rb
lymereJ: Also, I'm not sure if this needs to be specific to climate zone 3. There might be other climate zone sets having similar issues. What do you think?

lymereJ (commented):
I left a few comments, mostly about formatting but I do have one concerns about the main changes. We probably also need to update the envelope requirement lookup test to make sure that the lookup indeed works.

lib/openstudio-standards/standards/Standards.Model.rb
weilixu: That is correct. However, the specific sub climate zone info is lost after [model_get_climate_zone_set_from_list ](https://github.com/NREL/openstudio-standards/blob/9b2e994bd77651482e0a444cb2f17009b941c468/lib/openstudio-standards/standards/Standards.Model.rb#L4292) . 

The scenario is possible, One modification I could suggest is not to call the model_get_climate_zone_set_from_list function. Instead: 

For example:
props = model_find_object(standards_data['construction_properties'],
                                'template' => template,
                                'climate_zone_set' => climate_zone_set.min,
                                'intended_surface_type' => intended_surface_type,
                                'standards_construction_type' => standards_construction_type,
                                'building_category' => building_category)

if !prop:
props = model_find_object(standards_data['construction_properties'],
                                'template' => template,
                                'climate_zone_set' => climate_zone_set.max,
                                'intended_surface_type' => intended_surface_type,
                                'standards_construction_type' => standards_construction_type,
                                'building_category' => building_category)

lib/openstudio-standards/standards/Standards.Model.rb
weilixu: Possibly. For now it only addresses for Climate Zone 3 because 2004/PRM requirements. The future development could add more similar cases.

weilixu:
I am closing this pull request since the two proposed solutions require a complete overhaul.

PR #1255, opened 2022-01-18
Implement new workflow to search constructions.

weilixu:
The construction search in 90.1-2004 depends on min & max percent of surface.
The fix introduces a new workflow that added a new method called model_get_percent_of_surface_range, and it enables flexibility in identifying construction based on window to wall ratio by adding two search criteria: minimum_percent_of_surface and max_percent_of_surface.

The default behavior of this function returns nil for both min&max percent of surface. The detail implementation shall be overriden in the each individual routine (90.1-2004).

Add 90.1-2004 to test_small_office test set.

Tests has run locally for All ASHRAE prototype & baseline.

### DIFF
--- a/lib/openstudio-standards/prototypes/common/objects/Prototype.Model.rb
+++ b/lib/openstudio-standards/prototypes/common/objects/Prototype.Model.rb
@@ -2692,6 +2692,42 @@ Standard.class_eval do
     return true
   end
 
+  # Calculate a model's window or WWR
+  # Disregard space conditioning (assume all spaces are conditioned)
+  # which is true for most of not all prototypes
+  #
+  # @param model [OpenStudio::Model::Model] OpenStudio model object
+  # @param wwr [Boolean]
+  # @return [Numeric] Returns window to wall ratio (percentage) or window area.
+  def model_get_window_area_info(model, wwr = true)
+
+    window_area = 0
+    wall_area = 0
+
+    model.getSpaces.each do |space|
+      # Get zone multipler
+      multiplier = space.thermalZone.get.multiplier
+      space.surfaces.each do |surface|
+        next if surface.surfaceType != 'Wall'
+        next if surface.outsideBoundaryCondition != 'Outdoors'
+
+        # Get wall and window area
+        wall_area += surface.grossArea * multiplier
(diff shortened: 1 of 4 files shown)


test/doe_prototype/test_small_office.rb
```diff
@@ -3,7 +3,7 @@
 
 class TestSmallOffice < CreateDOEPrototypeBuildingTest
   building_types = ['SmallOffice']
-  templates = ['DOE Ref Pre-1980','DOE Ref 1980-2004','90.1-2007','90.1-2013','90.1-2016','90.1-2019']
+  templates = ['DOE Ref Pre-1980','DOE Ref 1980-2004','90.1-2004','90.1-2007','90.1-2013','90.1-2016','90.1-2019']
```
lymereJ, line 6: Let's maybe remove `90.1-2004` from the list so `templates` is consistent with the other building types.

lib/openstudio-standards/prototypes/common/objects/Prototype.Model.rb
```diff
@@ -2692,6 +2692,42 @@ def model_set_building_north_axis(model, north_axis)
...
   end
 
+  # Calculate a model's window or WWR
+  # Disregard space conditioning (assume all spaces are conditioned)
+  # which is true for most of not all prototypes
+  #
+  # @param model [OpenStudio::Model::Model] OpenStudio model object
+  # @param wwr [Boolean]
+  # @return [Numeric] Returns window to wall ratio (percentage) or window area.
+  def model_get_window_area_info(model, wwr = true)
```
lymereJ, line 2702: I'm fine with this, however, I wonder if this shouldn't be a wrapper for the `model_get_window_area_info_for_orientation` method we have in the `AppendixG_Dev` branch. It could have avoid redundant code, although this is probably faster. I'm fine either way, just wanted to point it out.

lymereJ (approved):

test/doe_prototype/test_small_office.rb
weilixu: 90.1-2004 is removed.

mdahlhausen (approved):

PR #1264, opened 2022-02-16
Add a check if capacity of the coil is larger than the database, we s…

weilixu:
The fix in this PR checks the maximum capacity of water source heat pump database to make sure no warnings issued due to search fails from excessive coil capacity.

### DIFF
--- a/lib/openstudio-standards/standards/Standards.CoilCoolingWaterToAirHeatPumpEquationFit.rb
+++ b/lib/openstudio-standards/standards/Standards.CoilCoolingWaterToAirHeatPumpEquationFit.rb
@@ -115,16 +115,14 @@ class Standard
 
     # Check to make sure properties were found
     if coil_props.nil?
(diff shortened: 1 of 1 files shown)


lymereJ (approved):

lymereJ:
@mdahlhausen - Have you had a chance to look at the proposed changes? Thanks!

mdahlhausen (commented):
I don't like the idea of looping through the data json each time to get the maximum capacity listed each time we need to get an efficiency value.  Is a hardcoded value sufficient?  And isn't the warning message that it can't find an efficiency value nearly always indicative of the coil size being too large?  Maybe we just add a note to the warning message that capacity may be too large.

weilixu:
Updated the code to use the size of the returned object list to decide which warning message to print if prop_coil is nil.

mdahlhausen (approved):

PR #1265, opened 2022-02-18
nrcan_350: Add option to choose between in-suite and central DOAS systems; Change space-type of the office room in MURBs

srgilani:


khaddad:
The test case ' NECB_insuite_central_doas_Tests' makes use of several standard methods:  'apply_weather_data', 'apply_loads', 'apply_envelope', 'apply_fdwr_srr_daylighting', 'apply_auto_zoning', 'apply_systems_and_efficiencies'. Chris Kirney recently suggested that we use a single method to apply all these methods. The standard method 'model_apply_standard' can be used to perform all these steps. We have so far been using the current approach for many test cases. Chris and I have started using the single call method approach and I suggest you do the same here.

PR #1275, opened 2022-03-01
Adjust fan power based on zone multiplier.

weilixu:
Address the large fan power for VAV_med in large office building.
Adjust fan power based on zone multiplier


weilixu:
Close this PR since the logic need to be reviewed again.

PR #1301, opened 2022-04-11
check vav part load curve

wsjsxn:
Check if correct  VAV fan part load curves from appendix g Table G3.1.3.15 is applied to baseline.

### DIFF
--- a/test/90_1_prm/data/prototype_list.json
+++ b/test/90_1_prm/data/prototype_list.json
@@ -29,13 +29,9 @@
         ["Warehouse", "90.1-2013", "ASHRAE 169-2013-2A", "no_user_data",[]]
     ],
     "vav_fan_curve": [
-        ["Outpatient", "90.1-2013", "ASHRAE 169-2013-2A", "no_user_data",[]],
         ["MediumOffice", "90.1-2013", "ASHRAE 169-2013-2A", "no_user_data",[["remove_transformer", []]]],
-        ["LargeHotel", "90.1-2013", "ASHRAE 169-2013-2A", "no_user_data",[["remove_transformer", []]]],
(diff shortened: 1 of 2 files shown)


test/90_1_prm/data/prototype_list.json
```diff
@@ -28,6 +28,15 @@
         ["SmallOffice", "90.1-2013", "ASHRAE 169-2013-2A", "no_user_data",[]],
         ["Warehouse", "90.1-2013", "ASHRAE 169-2013-2A", "no_user_data",[]]
     ],
+    "vav_fan_curve": [
```
dmaddoxwhite, line 31: We don't need this many tests. I would use MediumOffice in CZ 4A and also 2A, and SecondarySchool in 4A only. That gets PVAV, VAV, and also includes a case with parallel fan powered boxes.

test/90_1_prm/test_appendix_g_prm.rb
```diff
@@ -1701,6 +1701,54 @@ def check_number_of_cooling_towers(prototypes_base)
...
+
+      
+        # check fan curves
+        # Skip single-zone VAV fans
+        next if supply_fan.airLoopHVAC.get.thermalZones.size == 1
+        # coefficient 1
+        if supply_fan.fanPowerCoefficient1.is_initialized
+          expected_coefficient = 0.0013
+          coefficient = supply_fan.fanPowerCoefficient1.get
+          assert((coefficient - expected_coefficient).abs < 0.01, "Expected Coefficient 1 for #{supply_fan_name} to be equal to #{expected_coefficient}; found #{coefficient} instead")
```
dmaddoxwhite, line 1720: The test tolerance of 0.01 is too large. You could fix it by doing a percentage calculation:
((coefficient - expected_coefficient)/expected_coefficient).abs < 0.01

dmaddoxwhite (approved):

Issue #274, opened 2017-10-04
create_prm_baseline_building should remove existing daylight controls


eringold:
This: https://github.com/NREL/openstudio-standards/blob/master/openstudio-standards/lib/openstudio-standards/standards/Standards.Model.rb#L169

should be 
>added = space.add_daylighting_controls(template, __true__, false)

to remove any daylight controls in the proposed model beyond those required by code.

MatthewSteen:
Now here...

https://github.com/NREL/openstudio-standards/blob/5fe0f175d9156c567dfe83a4c4c9948bb65e616c/lib/openstudio-standards/standards/Standards.Model.rb#L1877

mdahlhausen:
Fixed with #1303

PR #1303, opened 2022-04-15
remove existing daylighting controls

mdahlhausen:
remove existing daylighting controls by default for prototype and prm workflow methods

addresses issue [#274](https://github.com/NREL/openstudio-standards/issues/274)


mdahlhausen:
@weilixu can you take a look at the test failures?  Failures are only in a specific PRM test and I think it may be related to Issue 1 in #1393 

`Model created, but had Errors: ERROR - [openstudio.model.ThermalZone] Fraction of Zone Controlled by Secondary Daylight Control is 0.127463 and you supplied a Primary Fraction of 0.964 which would result in a sum greater than 1.0,ERROR - [openstudio.model.ThermalZone] Fraction of Zone Controlled by Primary Daylight Control is 1 and you supplied a Secondary Fraction of 0.035 which would result in a sum greater than 1.0`

Issue #1119, opened 2021-06-30
Add constant schedule ruleset value tolerance


agonzalezesteve:
`model_add_constant_schedule_ruleset` method should have a tolerance with value because when having a temperature with periodic value, for example OpenStudio.convert(180, "F", "C").get, adds lots of schedules.

https://github.com/NREL/openstudio-standards/blob/c855838717a078e877789a7206865cbb311548fa/lib/openstudio-standards/standards/Standards.Model.rb#L1952

For example using `(existing_day_sch_vals[0] - value).abs < 1e-6` instead of `existing_day_sch_vals[0] == value`

mdahlhausen:
fixed with #1304

PR #1304, opened 2022-04-15
change constant ruleset value checking

mdahlhausen:
change existing schedule value check from an exact comparison to within a small tolerance

addresses issue #1119


mdahlhausen:
@lymereJ this pull requests adds a tolerance value check instead of equality check when adding constant ruleset schedules.  It prevents adding duplicate copies identical of constant day schedules like 'Mixed Water At Faucet Temp - 140F Default 17' which are prolific throughout the prototypes.  Currently the model regression tests are failing because the duplicate schedules aren't generated.  Could you regenerate the regression models?

lymereJ:
@mdahlhausen - I just uploaded them.

mdahlhausen:
@lymereJ thanks!

PR #1308, opened 2022-04-20
Economizer exception prm

weilixu:


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.AirLoopHVAC.rb
```diff
@@ -15,6 +15,130 @@ def air_loop_hvac_unoccupied_threshold
...
+  #
+  # @param air_loop_hvac [OpenStudio::Model::AirLoopHVAC] air loop
+  # @param climate_zone [String] ASHRAE climate zone, e.g. 'ASHRAE 169-2013-4A'
+  # @return [Bool] returns true if required, false if not
+  def air_loop_hvac_prm_baseline_economizer_required?(air_loop_hvac, climate_zone)
+    economizer_required = false
+    baseline_system_type = air_loop_hvac.additionalProperties.getFeatureAsString('baseline_system_type').get
+    # System type 3 through 8 and 11, 12 and 13
+    if ['SZ_AC', 'PSZ_AC', 'PVAV_Reheat', 'VAV_Reheat', 'SZ_VAV', 'PSZ_HP', 'SZ_CV', 'PSZ_HP', 'PVAV_PFP_Boxes', 'VAV_PFP_Boxes'].include? baseline_system_type
+      case climate_zone
```
dmaddoxwhite, line 85: I think we should check for climate zone match more generically, in case the climate zone sets change in the future. Start with a list of strings to check like econo_not_required = ['0A', '0B', '1A', '1B', '2A', '3A', '4A'].
Then loop through this list and check for a string match in the current climate_zone. I'm not sure that CZs 7 and 8 will always have an "A" and "B" vesion, so checking the not_required list is probably safer.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.AirLoopHVAC.rb
```diff
@@ -15,6 +15,130 @@ def air_loop_hvac_unoccupied_threshold
...
+  # @param air_loop_hvac [OpenStudio::Model::AirLoopHVAC] air loop
+  # @param climate_zone [String] ASHRAE climate zone, e.g. 'ASHRAE 169-2013-4A'
+  # @return [Array<Double>] [economizer_type, drybulb_limit_f, enthalpy_limit_btu_per_lb, dewpoint_limit_f]
+  def air_loop_hvac_prm_economizer_type_and_limits(air_loop_hvac, climate_zone)
+    economizer_type = 'NoEconomizer'
+    drybulb_limit_f = nil
+    enthalpy_limit_btu_per_lb = nil
+    dewpoint_limit_f = nil
+
+    case climate_zone
```
dmaddoxwhite, line 30: I suggest to create lists of CZs for each category and loop them to assign the value (see similar comment below). There is no 65 F category for Appendix G, all the missing CZs from the high-limit table do not get economizer.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.AirLoopHVAC.rb
```diff
@@ -15,6 +15,130 @@ def air_loop_hvac_unoccupied_threshold
...
+        if air_loop_hvac.additionalProperties.getFeatureAsString('zone_group_type').get == 'computer_zones'
+          economizer_required = false
+        end
+      end
+    end
+
+    # Check user_data in the zones
+    # As discussed for now, if a zone under the air loop hvac system has economizer exception
+    # The HVAC economizer should not be required.
+    # Another ticket may need to add to create a separate HVAC air loop for
```
dmaddoxwhite, line 124: I don't think we need to comment on this here. No change would be needed in this method if we apply the exceptions to the system group method. I've added this issue to a Jira ticket so we don't lose track of it.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.Model.rb
```diff
@@ -336,9 +336,36 @@ def handle_multi_building_area_types(model, climate_zone, default_hvac_building_
       wateruse_equipment.additionalProperties.setFeature('building_type_for_swh', type_for_swh)
     end
 
+    # ============================Process airloop info ============================================
```
dmaddoxwhite, line 339: As we discussed, move this to a new method, called from handle_user_data(). The method handle_user_data() is a new method which will also call handle_multi_building_area_types(), as well as future user data related processing.

test/90_1_prm/test_appendix_g_prm.rb
```diff
@@ -1753,6 +1753,36 @@ def check_number_of_cooling_towers(prototypes_base)
     end
   end
 
+  def check_economizer_exception(baseline_base)
+    baseline_base.each do |baseline, baseline_model|
+      building_type, template, climate_zone, user_data_dir, mod = baseline
+      baseline_model.getAirLoopHVACs.each do |air_loop|
+        economizer_activated_target = false
+        temperature_hihglimit_target = 23.89
```
dmaddoxwhite, line 1761: Typo in highlimit vs. hihglimit

dmaddoxwhite (commented):
Looks good overall. See my specific comments.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.AirLoopHVAC.rb
weilixu: Updated https://github.com/NREL/openstudio-standards/blob/economizer_exception_prm/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.AirLoopHVAC.rb#L53

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.AirLoopHVAC.rb
weilixu: Updated: https://github.com/NREL/openstudio-standards/blob/economizer_exception_prm/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.AirLoopHVAC.rb#L31

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.Model.rb
weilixu: Updated: https://github.com/NREL/openstudio-standards/blob/economizer_exception_prm/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb#L655

weilixu:
Local tests passed.
Running CI - once CI passed, will do the merge.

PR #1312, opened 2022-04-22
Set lpd from userdata

weilixu:
Multi type (space, space type) handling for baseline lighting power density.
PR include implementation and test cases


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb
```diff
@@ -0,0 +1,396 @@
...
+        space_volume = space.volume
+        space_area = space.floorArea
+        space_height = OpenStudio.convert(space_volume / space_area, 'm', 'ft').get
+        # calculate and add new lpd values
+        user_space_type_lighting_area = (lighting_per_length * space_height +
+          lighting_per_area) * sub_space_type_frac
+        space_lighting_per_area += user_space_type_lighting_area
+      end
+      # Now calculate the occupancy control credit factor (weighted by frac_lpd)
+      occupancy_control_credit_sum += occ_control_reduction_factor * user_space_type_lighting_area * sub_space_type_frac
```
dmaddoxwhite, line 390: I think this also needs to be weighted by space_lighting_per_area.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb
```diff
@@ -0,0 +1,396 @@
...
+        OpenStudio.logFree(OpenStudio::Info, 'openstudio.standards.SpaceType', "Removed #{inst.name} from #{space_type.name}.")
+        inst.remove
+      end
+    end
+
+    # Get userdata from userdata_space and userdata_spacetype
+    user_spaces = @standards_data.key?('userdata_space') ? @standards_data['userdata_space'] : nil
+    user_spacetypes = @standards_data.key?('userdata_spacetype') ? @standards_data['userdata_spacetype'] : nil
+    if user_spaces && user_spaces.length >= 1
+      # call this function to enforce space-space_type one on one relationship
```
dmaddoxwhite, line 60: In the space type section below, you are checking has_multi_lpd_values_user_data before proceding. Wouldn't we want to do that here as well? If the user data has wwr building types but not lighting space types, then we don't need to call space_to_space_type_apply_lighting.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb
```diff
@@ -0,0 +1,396 @@
...
+  # @param user_spaces hash data contained in the user space
+  # @param user_spacetypes hash data contained in the user spacetypes
+  # @param space_type OpenStudio::Model::SpaceType object
+  def space_to_space_type_apply_lighting(user_spaces, user_spacetypes, space_type)
+    space_lighting_per_area_hash = {}
+    # first priority - user_space data
+    if user_spaces && user_spaces.length >= 1
+      space_type.spaces.each do |space|
+        user_space_index = user_spaces.index { |user_space| user_space['name'] == space.name.get }
+        unless user_space_index.nil?
```
dmaddoxwhite, line 123: It's possible that some spaces would have user data at the space level, but not lighting space type userdata (i.e. only WWR bulding type). So instead of "unless user_space_index.nil?", I think this should be something like:

unless user_space_index.nil? num_std_ltg_types = user_data['num_std_ltg_types'].to_i
if num_std_ltg_types > 0

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb
```diff
@@ -0,0 +1,396 @@
...
+          space_lighting_per_area = calculate_lpd_from_userdata(user_space_data, space)
+          space_lighting_per_area_hash[space.name.get] = space_lighting_per_area
+        end
+      end
+    end
+    # second priority - user_spacetype
+    if user_spacetypes && user_spacetypes.length >= 1
+      # if space type has user data
+      user_space_type_index = user_spacetypes.index { |user_spacetype| user_spacetype['name'] == space_type.name.get }
+      unless user_space_type_index.nil?
```
dmaddoxwhite, line 134: Similar comment here about "unless user_space_type_index.nil?". Currently space type user data only has lighting space types, but that might change in the future.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb
weilixu: The `space_lighting_per_area` is included at line [390](https://github.com/NREL/openstudio-standards/blob/set_lpd_from_userdata/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb#L393)

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb
weilixu: Added a function has_user_lpd_values as conditional test.
The function checks if the user space / spaceType has lighting data - if test failed, then it moves to next priority

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb
weilixu: added a check

test/90_1_prm/test_appendix_g_prm.rb
```diff
@@ -397,6 +397,51 @@ def check_envelope(prototypes_base)
...
+    prototypes_base.each do |prototype, model_baseline|
+      building_type, template, climate_zone, user_data_dir, mod = prototype
+      if user_data_dir == 'no_user_data'
+        sub_prototypes_base = {}
+        sub_prototypes_base[prototype] = model_baseline
+        check_lpd(sub_prototypes_base)
+      else
+        if user_data_dir == 'userdata_lpd_01'
+          space_name_to_lpd_target = {}
+          space_name_to_lpd_target['Attic'] =15.06948107
```
dmaddoxwhite, line 413: Do you have a markdown file for documentation? Maybe we can document the basis for the expected values in the markdown file. That way if they fail in the future due to some change in either the prototypes or the standards, we can more easily debug the failures.

dmaddoxwhite (commented):
I don't see a markdown file. Have you created one?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb
dmaddoxwhite: I'm a bit confused about lights_have_info. Should the occupance_control_credit_sum be inside the "if lights_have_info" block? Otherwise, user_space_type_lighting_area might not have the correct value?
Also, should the name be changed from user_space_type_lighting_area to user_space_type_lighting_per_area?

dmaddoxwhite: Also, maybe we should give an error if frac_sum would exceed 1.0.

weilixu: > I'm a bit confused about lights_have_info. Should the occupance_control_credit_sum be inside the "if lights_have_info" block? Otherwise, user_space_type_lighting_area might not have the correct value? Also, should the name be changed from user_space_type_lighting_area to user_space_type_lighting_per_area?

The `lights_have_info` is copied from legacy code that checks if the data has `lighting_per_area` or `lighting_per_length`. This particularly [block](https://github.com/NREL/openstudio-standards/blob/set_lpd_from_userdata/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb#L394-L403) was created to contain the calculation of `user_type_lighting_per_area` from `lighting_per_area` and `lighting_per_length`. The calculation of `occupancy_control_credit_sum` is left outside the block but it wont be affected by the logic inside the block because we have initialized the parameters [here](https://github.com/NREL/openstudio-standards/blob/set_lpd_from_userdata/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb#L385).

However, what I think may make sense is to move this [part](https://github.com/NREL/openstudio-standards/blob/set_lpd_from_userdata/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb#L387-L392) to the `occupancy_control_credit_sum` section so the code may become more readable?

weilixu: > Also, maybe we should give an error if frac_sum would exceed 1.0.

By error you mean an error that should go to log but won't affect the execution, or the error that will halt the execution? (I think should be the second option?)
Also, that triggers another question:
What if the user data does not add up to 100%?
1. We can either raise exception or
2. We use other data to fill up the rest?

test/90_1_prm/test_appendix_g_prm.rb
weilixu: Will add this in the markdown file, a new section called Test Cases.

weilixu:
@dmaddoxwhite I moved code from prm_2019 file to prm.
If CI passed, could you do a review again?
The electrical equipment ticket has dependency on it.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.SpaceType.rb
weilixu: Conclusion for exceed or under 1.0.
Both case shall be failed in a pre-processing check (new jira task)

dmaddoxwhite (approved):
Looks good!

PR #1319, opened 2022-05-25
Set plug load from userdata

weilixu:


weilixu:
Added space and spacetype processing in plug load

dmaddoxwhite (approved):
Looks great! I hand-checked all of your test cases, and they match expected values.

PR #1332, opened 2022-06-13
Handle baseline systems return air type

lymereJ:
The proposed changes determine the baseline systems return air type based on the proposed model configuration. While this is not explicitly required by Appendix G, it is a recommended procedure in the  ASHRAE 90.1-2019 PRM Reference Manual.


lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -7456,6 +7470,148 @@ def model_set_central_preheat_coil_spm(model, thermalZones, coil)
...
+          else
+            zone_return_air_type[return_air_type] = zone_design_air_flow
+          end
+
+          if zone.additionalProperties.hasFeature('plenum')
+            plenum = zone.additionalProperties.getFeatureAsString('plenum').get
+
+            if zone_return_air_type.keys.include?('plenum')
+              if zone_return_air_type['plenum'].keys.include?(plenum)
+                zone_return_air_type['plenum'][plenum] = zone_design_air_flow
```
dmaddoxwhite, line 7580: Should line 7580 be using += to accumulate the air flow values?

lymereJ: Yup, good catch! Thanks.

PR #1335, opened 2022-06-29
Apxg dcv

leijerry888:
Implement DCV related ASHRAE 90.1 2019 Appendix G baseline model creation logic. Details are in [lib/openstudio-standards/standards/ashrae_90_1_prm/docs/dcv.md](https://github.com/NREL/openstudio-standards/compare/AppendixG_Dev...apxg_dcv?expand=1#diff-2029ab4f362aa95668a9e309b67fe26215990a8ce322c88bd8d8a51dd053c2e3)


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.AirLoopHVAC.rb
```diff
@@ -320,6 +320,116 @@ def air_loop_hvac_allowable_system_brake_horsepower(air_loop_hvac)
...
+    # check the following conditions at zone level
+    # zone > 500 sqft AND design occ > 25 ppl/ksqft
+
+    area_served_m2 = 0
+    num_people = 0
+    thermal_zone.spaces.each do |space|
+      area_served_m2 += space.floorArea
+      num_people += space.numberOfPeople
+    end
+    area_served_ft2 = OpenStudio.convert(area_served_m2, 'm^2', 'ft^2').get
```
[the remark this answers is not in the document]
dmaddoxwhite: Actually, disregard this. The occ per ksf is an average of the spaces.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -914,6 +914,225 @@ def model_set_central_preheat_coil_spm(model, thermalZones, coil)
...
+  # generation
+  #
+  # @author Xuechen (Jerry) Lei, PNNL
+  # @param model [OpenStudio::Model::Model] Openstudio model
+  def model_raise_user_model_dcv_errors(model)
+    model.getThermalZones.each do |thermal_zone|
+      if thermal_zone.additionalProperties.getFeatureAsBoolean('zone DCV implemented in user model').get &&
+         (!thermal_zone.additionalProperties.getFeatureAsBoolean('zone dcv required by 901').get ||
+           !thermal_zone.additionalProperties.getFeatureAsBoolean('airloop dcv required by 901').get)
+        OpenStudio.logFree(OpenStudio::Warn, 'openstudio.standards.Model', "For thermal zone #{thermal_zone.name}, ASHRAE 90.1 2019 6.4.3.8 does NOT require this zone to have demand control ventilation, but it was implemented in the user model, Appendix G baseline generation will continue!")
```
dmaddoxwhite, line 1070: I think this is not really interesting if we don't know whether DCV will be required in the baseline. If we move this test after model_add_apxg_dcv_properties(), we can also check "apxg no need to have DCV" for each zone. If NOT "no need", then we have a case where DCV is not required by chapter 6, but it is included in both the user model and the baseline. Then we can report that in the logFree message.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -914,6 +914,225 @@ def model_set_central_preheat_coil_spm(model, thermalZones, coil)
...
+      end
+      if thermal_zone.additionalProperties.getFeatureAsBoolean('zone dcv required by 901').get &&
+         thermal_zone.additionalProperties.getFeatureAsBoolean('airloop dcv required by 901').get &&
+         !thermal_zone.additionalProperties.getFeatureAsBoolean('zone DCV implemented in user model').get
+        OpenStudio.logFree(OpenStudio::Error, 'openstudio.standards.Model', "For thermal zone #{thermal_zone.name}, ASHRAE 90.1 2019 6.4.3.8 requires this zone to have demand control ventilation, but it was not implemented in the user model, Appendix G baseline generation should be terminated!")
```
dmaddoxwhite, line 1075: For this test, we should report the error, but not halt the generation. We should report to the special PRM log file using 'prm.log' instead of 'openstudio.standards.Model'. You can pull Appendix_G_Dev to get that capability into your branch. I think we should have a standard prefix for cases where the user model does not meet mandatory requirements.  Like start with: 'Mandatory Requirement Not Met:  '. Then the user can filter or sort the prm.log file to see those issues together.  I don't think the phrase "Appendix G baseline generation should be terminated" is meaningful if we are not actually terminating.

test/90_1_prm/test_appendix_g_prm.rb
```diff
@@ -733,6 +733,254 @@ def check_sizing_values(model, building_type, template, climate_zone)
...
+          zone = model_baseline.getThermalZoneByName('Cafeteria_ZN_1_FLR_1 ZN').get
+          airloop = zone.airLoopHVAC.get
+          assert(!dcv_is_on(zone, airloop))
+
+        when 7
+          # test 7
+          #   - DCV should NOT be in the user model (zone ppl density < 25 ppl/ksqft)
+          #   - DCV should NOT be in the baseline model (air loop oa flow < 3000 cfm || zone ppl density < 100 ppl/ksqft)
+          #   - DCV is implemented in user model
+          # expected result: baseline implements DCV but prompts warning (user model has DCV but meet exception)
```
dmaddoxwhite, line 871: I think this should say: "expected result: no DCV in baseline model".

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.AirLoopHVAC.rb
```diff
@@ -320,6 +320,116 @@ def air_loop_hvac_allowable_system_brake_horsepower(air_loop_hvac)
...
+  end
+
+  # Check if the thermal zone in baseline model needs to have DCV
+  #
+  # @author Xuechen (Jerry) Lei, PNNL
+  # @param thermal_zone [OpenStudio::Model::ThermalZone] the thermal zone
+  # @return [Boolean] flag of whether thermal zone in baseline is required to have DCV
+  def baseline_thermal_zone_demand_control_ventilation_required?(thermal_zone)
+    # zone needs dcv if user model has dcv and baseline does not meet apxg exception
+    if thermal_zone.additionalProperties.hasFeature("apxg no need to have DCV")
```
dmaddoxwhite, line 406: I believe this should be based on "zone dcv required by 901". Eventually we will have a Proposed model, and that model will follow 90.1 requirements, even if the user model does not have DCV. Same for second .HasFeature below.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -914,6 +914,225 @@ def model_set_central_preheat_coil_spm(model, thermalZones, coil)
...
+  # Set DCV in baseline HVAC system if required
+  #
+  # @author Xuechen (Jerry) Lei, PNNL
+  # @param model [OpenStudio::Model::Model] Openstudio model
+  def model_set_baseline_demand_control_ventilation(model, climate_zone)
+    model.getAirLoopHVACs.each do |air_loop_hvac|
+      if baseline_air_loop_hvac_demand_control_ventilation_required?(air_loop_hvac)
+        air_loop_hvac_enable_demand_control_ventilation(air_loop_hvac, climate_zone)
+        air_loop_hvac.thermalZones.sort.each do |zone|
+          unless baseline_thermal_zone_demand_control_ventilation_required?(zone)
```
dmaddoxwhite, line 1128: Also need to account for situation where user model does not have DCV, but it is required in the Proposed and Baseline models. If the user model does not have OA specified as per person, then we need to report and error. In this case I believe it should terminate the program, since the user can then correct the problem and re-run.

test/90_1_prm/test_appendix_g_prm.rb
```diff
@@ -733,6 +733,254 @@ def check_sizing_values(model, building_type, template, climate_zone)
...
+          airloop = zone.airLoopHVAC.get
+          # check warning
+          assert(dcv_is_on(zone, airloop))
+
+        when 3
+          # test case 3:
+          #   - DCV should be in the user model (zone ppl density > 25 ppl/ksqft, no user exception)
+          #   - DCV should be in the baseline model (air loop oa flow > 3000 cfm && zone ppl density > 100 ppl/ksqft)
+          #   - DCV is NOT implemented in user model
+          # expected result: error and terminate
```
dmaddoxwhite, line 798: Only terminate if zone OA requirement is not specified as flow per person. Otherwise, apply DCV to baselne.

PR #1336, opened 2022-06-29
PRM log system

weilixu:
Appendix G PRM log system

This PR is for discussion only. It has the implementation for a log system + a way to elegantly handle the PRM-specific errors.

### DIFF
--- a/lib/openstudio-standards/utilities/logging.rb
+++ b/lib/openstudio-standards/utilities/logging.rb
@@ -36,6 +36,54 @@ def log_messages_to_runner(runner, debug = false)
   end
 end
 
+# A function that writes error message to log file and stop the execution.
+#
+# @param msg [String] error message
+# @param log_level [OpenStudio::LogLevel] log level, eg. OpenStudio::Info
+# @param log_path [String] the log file directory
+# @param debug [Boolean] debug mode
+def terminate_prm_write_log(msg, log_level, log_path, debug = false)
+  OpenStudio.logFree(log_level, 'openstudio.prm', msg)
+  log_messages_to_file_prm(log_path, debug)
+  raise msg
+end
+
+# A function to export log messages specific to PRM to a log file.
+#
+# @param file_path [String] the file path to the log file
+# @param debug [Boolean] debug mode
+# @return message [Array] array of message strings
+def log_messages_to_file_prm(file_path, debug = false)
+  messages = []
+
+  File.open(file_path, 'w') do |file|
+    $OPENSTUDIO_LOG.logMessages.each do |msg|
+      # only apply to prm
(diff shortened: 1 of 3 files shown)


lib/openstudio-standards/utilities/logging.rb
```diff
@@ -36,6 +36,54 @@ def log_messages_to_runner(runner, debug = false)
...
+  log_messages_to_file_prm(log_path, debug)
+  raise msg
+end
+
+# A function to export log messages specific to PRM to a log file.
+#
+# @param file_path [String] the file path to the log file
+# @param debug [Boolean] debug mode
+# @return message [Array] array of message strings
+def log_messages_to_file_prm(file_path, debug = false)
```
lymereJ, line 56: Is this method the same as `log_messages_to_file` except for the following block? https://github.com/NREL/openstudio-standards/blob/abae01541a336d08f736eb00fad9fba24fea4f23/lib/openstudio-standards/utilities/logging.rb#L51-L60

lymereJ:
Looks good!

lib/openstudio-standards/utilities/logging.rb
weilixu: No. the log_message_to_file method 
1. only records 'openstudio.prm' logChannel (which later we may need to write in 'openstudio.prm' in the logFree.
2. A slightly different log [format](https://github.com/NREL/openstudio-standards/blob/ddc56188c3cf3f82cdce13a061de6fe5097180b1/lib/openstudio-standards/utilities/logging.rb#L65) we can decide later.

weilixu:
To write in PRM generation log we should set log Channel to 'openstudio.prm'

Here is the example https://github.com/NREL/openstudio-standards/blob/ddc56188c3cf3f82cdce13a061de6fe5097180b1/lib/openstudio-standards/standards/Standards.Model.rb#L460

```ruby
# set logChannel to openstudio.prm so this message will be exported to OpenStudio PRM log.
OpenStudio.logFree(OpenStudio::Info, 'openstudio.prm', 'This message is used as test for log message function')
```

lib/openstudio-standards/utilities/logging.rb
weilixu: That is correct - log_message_to_file_prm is added to the last line of the `model_create_prm_any_baseline_building` method - we will likely to create a private method and override in the ashrae 90.1 prm class to separate this report from the other routines later.

There are several categories that `terminate_prm_write_log` can be applied to. One of the them is the user data validation, which is included in the example in this PR. We may need a clearer definition on what should raise with `terminate_prm_write_log` function, and i hope to use this PR as example to help with the discussion.

dmaddoxwhite (approved):
Looks good!

PR #1348, opened 2022-07-27
Apxg prm inheritance

dmaddoxwhite:


lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -3093,6 +2354,12 @@ def model_add_daylighting_controls(model)
     return true
   end
 
+  # For backward compatibility, infiltration standard not used for 2013 and earlier
+   # @return [Bool] true if successful, false if not
+  def baseline_apply_infiltration_standard(model, climate_zone)
```
lymereJ, line 2359: To be consistent with other methods, the name should probably be `model_baseline_apply_infiltration_standard` since it's under `Standards.Model.rb`

lymereJ: We might also want to return `true` to be consistent with the YARD comments.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -936,363 +923,6 @@ def model_prm_baseline_system_groups(model, custom)
...
-        OpenStudio.logFree(OpenStudio::Info, 'openstudio.standards.Model', "--- #{zone_names.join(', ')}")
-      end
-    end
-
-    return final_groups
-  end
-
   # Before deleting proposed HVAC components, determine for each zone if it has district heating
   # @return [Hash] of boolean with zone name as key
   def get_district_heating_zones(model)
```
lymereJ, line 928: To be consistent with other methods, the name should probably be `model_get_district_heating_zones` since it's under `Standards.Model.rb`

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -613,13 +584,24 @@ def model_num_stories_spanned(model, zones)
...
   end
 
+
+  # Add design day schedule objects for space loads, 
+  # not used for 2013 and earlier
+  # @author Xuechen (Jerry) Lei, PNNL
+  # @param model [OpenStudio::model::Model] OpenStudio model object
+  #
+  def model_apply_prm_baseline_sizing_schedule(model)
+    return
```
lymereJ, line 594: I think the convention is to return `true` by default and `false` if it failed.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -382,12 +356,9 @@ def model_create_prm_any_baseline_building(user_model, building_type, climate_zo
...
 
-      if /prm/i !~ template
-        # If there are any multi-zone systems, reset damper positions to achieve a 60% ventilation effectiveness minimum for the system
-        # following the ventilation rate procedure from 62.1
-        model_apply_multizone_vav_outdoor_air_sizing(model)
-
-      end
+      # If there are any multi-zone systems, reset damper positions to achieve a 60% ventilation effectiveness minimum for the system
+      # following the ventilation rate procedure from 62.1
+      model_apply_multizone_vav_outdoor_air_sizing(model)
```
lymereJ, line 361: Do we actually want this to occur for the 90.1-2019 PRM? I thought we did not find it a requirement for it or a mention of it in the PRM-RM.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -1513,16 +1143,17 @@ def model_prm_baseline_system_group_minimum_area(model, custom)
   #
   # @param model [OpenStudio::Model::Model] OpenStudio model object
   # @param climate_zone [String] ASHRAE climate zone, e.g. 'ASHRAE 169-2013-4A'
-  # @param area_type [String] Valid choices are residential, nonresidential, and heatedonly
-  # @param fuel_type [String] Valid choices are electric, fossil, fossilandelectric,
-  #   purchasedheat, purchasedcooling, purchasedheatandcooling
-  # @param area_ft2 [Double] Area in ft^2
-  # @param num_stories [Integer] Number of stories
+  # @param sys_group [data structure] Data structure defining list of zones for a single hvac building type
```
lymereJ, line 1146: Isn't it a Hash?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -1563,4 +1574,757 @@ def handle_zone_hvac_user_input_data(model)
...
+    return final_groups
+  end
+
+  # Alternate method for 2016 and later stable baseline
+  # Limits for each building area type are taken from data table
+  # Heating fuel is based on climate zone, unless district heat is in proposed
+  #
+  # @note Select system type from data table base on key parameters
+  # @param climate_zone [string] id code for the climate
+  # @param sys_group [data structure] Data structure defining list of zones for a single hvac building type
```
lymereJ, line 1969: Hash instead of data structure?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.ThermalZone.rb
```diff
@@ -149,4 +149,22 @@ def thermal_zone_get_fan_power_limitations(thermal_zone, is_energy_recovery_requ
     # assuming that all supply air passes through all devices
     return fan_pwr_adjustment_in_wc * dsn_zone_air_flow_cfm / 4131
   end
+
+  # Identify if zone has district energy for occ_and_fuel_type method
+  # @param themal_zone
+  # @return [string] with applicable DistrictHeating and/or DistrictCooling 
+  def get_zone_fuels_for_occ_and_fuel_type(zone)
```
lymereJ, line 156: To be consistent with other methods, the name should probably be `thermal_zone_get_district_heating_zones` since it's under ThermalZone.rb`

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -652,35 +642,32 @@ def model_zones_with_occ_and_fuel_type(model, custom)
...
     end
 
     return zones
   end
 
+  # for 2013 and prior, baseline fuel = proposed fuel
+  # @param themal_zone
+  # @return [string] with applicable DistrictHeating and/or DistrictCooling 
+  def get_zone_fuels_for_occ_and_fuel_type(zone)
+    zone_fuels = thermal_zone_fossil_or_electric_type(zone, 'custom')
```
lymereJ, line 659: Why are we use `'custom'` instead of `custom`, it looks like the original call used `custom`.

lib/openstudio-standards/standards/Standards.Model.rb
dmaddoxwhite: Right. The 90_1_prm version just returns "true".

lib/openstudio-standards/standards/Standards.Model.rb
dmaddoxwhite: The custom variable is not actually used in the thermal_zone_fossil_or_electric_type() method. So rather than dragging it through, I just put in a string. I'll change it to an empty string to make it more clear what is happening.

lib/openstudio-standards/standards/Standards.Model.rb
dmaddoxwhite: OK

lib/openstudio-standards/standards/Standards.Model.rb
dmaddoxwhite: I've moved this method to Standards.ThermalZone.rb to be consistent with the 90_1_prm version.

PR #1351, opened 2022-08-05
Wwr increase feature

weilixu:


lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -6114,8 +6157,79 @@ def model_get_unmet_load_hours(model)
...
+  # Readjusted the WWR for surfaces previously has no windows to meet the
+  # overall WWR requirement.
+  # This function shall only be called if the maximum WWR value for surfaces with fenestration is lower than 90% due to
+  # accommodating the total door surface areas
+  #
+  # @param residual_ratio: [Float] the ratio of residual surfaces among the total wall surface area with no fenestrations
+  # @param space [OpenStudio::Model:Space] a space
+  # @param model [OpenStudio::Model::Model] openstudio model
+  # @return [Bool] return true if successful, false if not
+  def readjust_surface_wwr(residual_ratio, space, model)
```
lymereJ, line 6229: Perhaps we should rename the method name to be `model_readjust_surface_wwr` to be consistent with other methods in this file.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -6844,6 +6958,60 @@ def air_loop_hvac_supply_air_temperature_reset_type(air_loop_hvac)
...
+  def get_wwr_of_a_surface(surface)
+    surface_area = surface.grossArea
+    surface_fene_area = 0.0
+    surface.subSurfaces.sort.each do |ss|
+      next unless ss.subSurfaceType == 'FixedWindow' || ss.subSurfaceType == 'OperableWindow' || ss.subSurfaceType == 'GlassDoor'
+
+      surface_fene_area += ss.netArea
+    end
+    return surface_fene_area / surface_area
+  end
```
lymereJ, line 6974: This should probably be moved to `Standards.Surface.rb` and perhaps be renamed `surface_get_wwr` or something similar to be consistent with other surface related methods.

Same for `get_door_ratio_of_a_surface`.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -6844,6 +6958,60 @@ def air_loop_hvac_supply_air_temperature_reset_type(air_loop_hvac)
...
+                              surface_wwr: 0.0,
+                              surface_dr: 0.0,
+                              wwr_building_type: 'All others',
+                              wwr_target: 0.0,
+                              total_wall_m2: 0.0,
+                              total_wall_with_fene_m2: 0.0,
+                              total_fene_m2: 0.0,
+                              total_plenum_wall_m2: 0.0)
+    return 1.0 - multiplier
+  end
```
lymereJ, line 7013: Perhaps we should rename the method name to be `model_get_wwr_reduction_ratio` to be consistent with other methods in this file.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -2003,6 +2003,155 @@ def handle_zone_hvac_user_input_data(model)
...
+        if surface_wwr == 0.0
+          # 1.0 means remain the original form
+          reduction_ratio = 1.0
+        else
+          reduction_ratio = multiplier
+        end
+      end
+    end
+    return reduction_ratio
+  end
```
lymereJ, line 2079: Same comments as for the methods in `Standards.Model.rb`

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -2003,6 +2003,155 @@ def handle_zone_hvac_user_input_data(model)
...
+    space.surfaces.sort.each do |surface|
+      next unless surface.additionalProperties.hasFeature('added_wwr')
+
+      added_wwr = surface.additionalProperties.getFeatureAsDouble('added_wwr').to_f
+      # The full calculation of adjustment is:
+      # ((residual_ratio * surface_area + added_wwr * surface_area) / surface_area ) / added_wwr
+      adjustment_ratio = residual_ratio / added_wwr + 1.0
+      model_adjust_fenestration_in_a_surface(surface, adjustment_ratio, model)
+    end
+  end
```
lymereJ, line 2154: Same comments as for the methods in `Standards.Model.rb`

test/90_1_prm/test_appendix_g_prm.rb
```diff
@@ -1908,7 +1912,7 @@ def check_multi_bldg_handling(baseline_base)
         wwr_baseline = run_query_tabulardatawithstrings(model_baseline, 'InputVerificationandResultsSummary', 'Conditioned Window-Wall Ratio', 'Gross Window-Wall Ratio', 'Total', '%').to_f
         # Check WWR against expected WWR
         wwr_goal = 100 * @@wwr_values[building_type].to_f
-        assert(wwr_baseline < wwr_goal, "Baseline WWR for the #{building_type}, #{template}, #{climate_zone} model with user data is incorrect. The WWR of the baseline model is #{wwr_baseline} but should be #{wwr_baseline}, smaller than the WWR goal #{wwr_goal}")
+        assert(wwr_baseline > wwr_goal, "Baseline WWR for the #{building_type}, #{template}, #{climate_zone} model with user data is incorrect. The WWR of the baseline model is #{wwr_baseline} but should be greater than the WWR goal #{wwr_goal}")
```
lymereJ, line 1915: Shouldn't we be looking for a difference instead? As in is `wwr_baseline` different than `wwr_goal`.

lib/openstudio-standards/standards/ashrae_90_1_prm/docs/set_baseline_wwr.md
```diff
@@ -0,0 +1,191 @@
...
+The function `get_wwr_reduction_ratio` implements the logic for calculating the reduction ratio `reduction_ratio` and `model_adjust_fenestration_in_a_surface` implements the logics to adjust the window to wall ratio for a surface
+
+The implementation of both functions can be found in the `ashrae_90_1_prm.Model.rb`
+
+## Notes:
+The function works well with cases where WWR reduction is needed. For WWR increase to meet Table G3.1.1-1 values, the function's limitation is listed below:
+1. The function will remove all windows in the surface then add a new window to meet the required WWR.
+2. The function will keep the existing doors but there could be cases the location of door overlaps with the new window.
+
+In addition, when increasing the WWR requires adding windows to surfaces with no windows, the function will not adding any windows to spaces tagged as plenum and used as a AirLoopSupplyPlenum or AirLoopReturnPlenum.
```
lymereJ, line 59: Should it be "to spaces tagged as plenum **and/or** used as a AirLoopSupplyPlenum or AirLoopReturnPlenum

lib/openstudio-standards/standards/ashrae_90_1_prm/docs/set_baseline_wwr.md
```diff
@@ -0,0 +1,191 @@
...
+
+# Code Requirement Interpretation
+1. Fenestration area shall be calculated based on the area of gross above-grade walls that separate conditioned spaces and semiheated spaces from the exterior
+2. If the window to wall ratio building type (wwr_building_type) can be found in Table G3.1.1-1, then the baseline window to wall ratio shall equal to the values specified in the table by proportionally decrease or increase the fenestration area.
+3. If the wwr_building_type cannot be found in the Table or it is other type, the baseline window to wall ratio shall equal to 40% or proposed window to wall ratio, whichever is smaller.
+4. Multiple wwr_building_types in a building shall use the values in the table correspondingly.
+5. If the fenestration area increase exceeds the host wall surfaces, then the residual shall be distributed to the other wall surfaces proportionally to its wall surface area
+6. Existing building shall remain the fenestration area the same as the existing fenestration area prior to the proposed work
```
lymereJ, line 17: Are we handling existing building using the user data? I may have missed it but I have not seen anything related to handling that feature for now.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -4494,20 +4494,28 @@ def model_apply_prm_baseline_window_to_wall_ratio(model, climate_zone, wwr_build
...
+        bat.store('total_plenum_m2', 0.001)
       else
         bat = bat_win_wall_info[std_spc_type]
       end
 
       # Loop through all surfaces in this space
       wall_area_m2 = 0
       wind_area_m2 = 0
+      wall_only_area_m2 = 0
+      num_wall_no_fene = 0
```
lymereJ, line 4518: Is this unused?

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -6114,8 +6157,79 @@ def model_get_unmet_load_hours(model)
     return result
   end
 
+  # A function to check whether a space is a return / supply plenum.
+  # This function only works on spaces that is categorized as return or supply air plenum
+  # For zones works as plenum but not correctly categorized in the OS will not be identified by this function
```
lymereJ, line 6162: Unless I am missing something, this function doesn't seem to look if a space has been categorized as a plenum but only looks the type of return plenums that the space be might used for.

lymereJ (commented):
Thanks for addressing this, this is going to be a great feature to have! I've added a comments/questions, some of them might need further discussions, and some are probably trivial.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -6114,8 +6157,79 @@ def model_get_unmet_load_hours(model)
...
+    # the zone is a supply plenum
+    space.model.getAirLoopHVACSupplyPlenums.each do |supply_air_plenum|
+      if supply_air_plenum.thermalZone.get.name.to_s == zone.get.name.to_s
+        return true
+      end
+    end
+    # None match, return false
+    return false
+  end
+
```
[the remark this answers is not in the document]
weilixu: Moved.

We discussed this might the solution for now and documented in the MD about its limitation.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -6114,8 +6157,79 @@ def model_get_unmet_load_hours(model)
...
+  # accommodating the total door surface areas
+  #
+  # @param residual_ratio: [Float] the ratio of residual surfaces among the total wall surface area with no fenestrations
+  # @param space [OpenStudio::Model:Space] a space
+  # @param model [OpenStudio::Model::Model] openstudio model
+  # @return [Bool] return true if successful, false if not
+  def readjust_surface_wwr(residual_ratio, space, model)
+    return true
+  end
+
```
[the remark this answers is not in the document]
weilixu: This is a good question. Lot times it is hard to know when to make a function public / private. For these two functions, I put them in private primarily because:

1. incomplete logic -> The `model_adjust_fenestration_in_a_surface` does not take care of reduction ratio >=0 case, this is because the original code has not considered these cases.
2. Arguments are specific - residual_ratio is the ratio of additional surfaces need to be added to the total net wall surface area. This function mainly handles when the total fenestration surface is over wall surface area, so we need to relocate some of the additional fenestration areas to those net wall surfaces.
3. Only used in one routine - the readjust_surface_wwr only used in 90.1 PRM.

However, it does not mean these two function should stay private. I would be happy to hear any use cases that user may found these two functions useful to access.

test/90_1_prm/test_appendix_g_prm.rb
weilixu: This was a bug that motivated me to work on the unit test...

It is actually the multi-bldg-handling check function. Background story is, this test case is a small office with three rooms assigned to three different types of WWR space types. So in total, there are four different WWR space types.
- Retail (stand alone) - 11%  - Perimeter_ZN_2
- Retail (strip mall) - 20% -  - Perimeter_ZN_1
- Restaurant (quick service) - 34% - Perimeter_ZN_4
- Office (<=5000 ft2) - 19%.
The `wwr_goal` is 19% because it is from the json file.

Previously, the WWR function can only shrink, so obviously, because of the Retail (stand alone), the total WWR should be smaller than 19%.

With the new code, Restaurant and Retail strip mall both have higher WWR and the total WWR is slightly over 20% so the generated `wwr_baseline` should be higher (>) than `wwr_goal`.

To be more precise, I can use (wwr_baseline - wwr_goal) == 1% but so far > operator might be OK. Let me know your thoughts.

lib/openstudio-standards/standards/ashrae_90_1_prm/docs/set_baseline_wwr.md
weilixu: I think I missed the implementation of this part... I will add it to the code in the next push

lib/openstudio-standards/standards/Standards.Model.rb
weilixu: Yes, the `num_wall_no_fene` is not used.
It was part of implementation on handling windows + door surface area but I quickly move to another implementation approach.
Must forget to delete it. Good catch, thanks.

lib/openstudio-standards/standards/Standards.Model.rb
weilixu: I will update the wording based on the one written in the document.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -4653,45 +4676,65 @@ def model_apply_prm_baseline_window_to_wall_ratio(model, climate_zone, wwr_build
           when 'Unconditioned'
             next # Skip unconditioned spaces
           when 'NonResConditioned'
-            next unless vals['red_nr']
-
             mult = vals['mult_nr_red']
+            total_wall_area = vals['nr_wall_m2']
+            total_wall_with_fene = vals['nr_fene_only_wall_m2']
```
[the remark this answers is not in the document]
weilixu: Thanks for the suggestion. I agree the `total_wall_area` should rename to avoid confusion. However, the `total_wall_with_fene` is not exactly the `total_net_wall_area`

`total_wall_with_fene` is the total area of the vertical surfaces that have fenestrations. It is still the gross area of walls. The difference is that the wall gross area is only added if the wall has fenestration surfaces. I am using this parameter to determine what is the maximum WWR a model can reach by only expanding the existing windows.

lib/openstudio-standards/standards/ashrae_90_1_prm/docs/set_baseline_wwr.md
weilixu: A follow up questions @lymereJ @dmaddoxwhite do you have preference on where to add this check?
I can do it at high-level such as somewhere at the beginning of the [function](https://github.com/NREL/openstudio-standards/blob/AppendixG_Dev/lib/openstudio-standards/standards/Standards.Model.rb#L4450-L4454)
- Pro: avoid all the calculations and API calls - save time especially for large models.
- Con: we might need to create a dummy function again in Standard.Model.rb and override it in the PRM routine.

I can also do it at sub [function](https://github.com/NREL/openstudio-standards/blob/wwr_increase_feature/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb#L2033-L2043) level.
- Pro: no structure change to parent class.
- Con: Will still require all the calculation and API calls.

lib/openstudio-standards/standards/Standards.Model.rb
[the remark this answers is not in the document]
lymereJ: I see, that makes sense! Would it be alright just adding the `_area` suffix to be consistent with the other variables?

lib/openstudio-standards/standards/Standards.Model.rb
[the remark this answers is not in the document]
lymereJ: Sounds good!

test/90_1_prm/test_appendix_g_prm.rb
lymereJ: That sounds good!

lib/openstudio-standards/standards/ashrae_90_1_prm/docs/set_baseline_wwr.md
lymereJ: Mmmh, good one. I think I would lean towards the first solution you proposed, mostly for the pro, but I don't have a strong opinion on it.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -4615,12 +4634,14 @@ def model_apply_prm_baseline_window_to_wall_ratio(model, climate_zone, wwr_build
...
+
       # Check against WWR limit
       vals['red_nr'] = vals['wwr_nr'] > wwr_lim
       vals['red_res'] = vals['wwr_res'] > wwr_lim
       vals['red_sh'] = vals['wwr_sh'] > wwr_lim
 
-      # Stop here unless windows need reducing or increasing if
+      # Stop here unless windows need reducing or increasing or if
       # following the stable baseline approach
       return true, base_wwr unless (vals['red_nr'] || vals['red_res'] || vals['red_sh']) || template == '90.1-PRM-2019'
```
dmaddoxwhite, line 4646: We've been trying to get away from using template as the trigger for stable baseline features. I think a good alternative might be to add a method named "stable_baseline?", and use inheritance to give it a value of "true" for 90_1_prm and "false" in Standards.Model.rb. Also applies to line 4619 above.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -4643,6 +4664,8 @@ def model_apply_prm_baseline_window_to_wall_ratio(model, climate_zone, wwr_build
         std_spc_type = space.additionalProperties.getFeatureAsString('building_type_for_wwr').get
```
dmaddoxwhite, line 4664: Will this block of code be used only for stable baseline? If not, will this additional properties fail if it is not a stable baseline?

lib/openstudio-standards/standards/ashrae_90_1_prm/docs/set_baseline_wwr.md
dmaddoxwhite: I agree with first solution.

weilixu: As discussed, we will keep the existing building feature for the next release.

lib/openstudio-standards/standards/Standards.Model.rb
weilixu: There is a [code block](https://github.com/NREL/openstudio-standards/blob/45193c51165428637294bc36463445486d2f8fb5/lib/openstudio-standards/standards/Standards.Model.rb#L4473-L4484) at the top of this function adds `building_type_for_wwr` to each space. So in here, every space should have this additional property.

dmaddoxwhite (approved):
Looks good!

PR #1356, opened 2022-08-16
Optimize sizing run in prm

weilixu:
PR for review.
Fixed issues:
1. Bug in rotation building code
2. Optimized sizing runs with the bug fix
3. Restructure the unit tests for PRM so the CI runs test in parallel


dmaddoxwhite (approved):
Looks good!

PR #1366, opened 2022-08-31
Handle baseline door constructions

yunyangye:
Fix the baseline door constructions issue. The json files, ruby script, and test file were updated accordingly.


dmaddoxwhite (approved):
Looks good!

PR #1373, opened 2022-09-08
Fixes #1372 by updating VentilationRateProcedure enumerations

### LINKED ISSUE #1372: VentilationRateProcedure enums changed
In OpenStudio 3.3.0/EnergyPlus 9.6, the enumerations in the `Sizing:System` and `Controller:MechanicalVentilation` objects changed from  `VentilationRateProcedure` to `Standard62.1VentilationRateProcedure`. When this change is not handled, the setter methods will not change the value, and the default `ZoneSum` method will be used. This method significantly underestimates the minimum outdoor air requirements for multizone systems, which in turn impacts energy consumption.

The fix is to modify the code to be version-dependent.

asparke2:
In OpenStudio 3.3.0/EnergyPlus 9.6, the enumerations in the Sizing:System and Controller:MechanicalVentilation objects changed from VentilationRateProcedure to Standard62.1VentilationRateProcedure. When this change is not handled, the setter methods will not change the value, and the default ZoneSum method will be used. This method significantly underestimates the minimum outdoor air requirements for multizone systems, which in turn impacts energy consumption.

The fix is to modify the code to be version-dependent.

### DIFF
--- a/lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb
+++ b/lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb
@@ -1811,7 +1811,7 @@ class Standard
         controller_mv.setSystemOutdoorAirMethod('Standard62.1VentilationRateProcedureWithLimit')
       end
         # Change the min flow rate in the controller outdoor air
-      controller_oa.setMinimumOutdoorAirFlowRate(0.0)
+      # controller_oa.setMinimumOutdoorAirFlowRate(0.0)
     else
       OpenStudio.logFree(OpenStudio::Warn, 'openstudio.standards.AirLoopHVAC', "For #{air_loop_hvac.name}, cannot enable multizone vav optimization because the system has no OA intake.")
       return false
@@ -1850,7 +1850,7 @@ class Standard
         if equip.to_AirTerminalSingleDuctVAVReheat.is_initialized
           zone_oa = thermal_zone_outdoor_airflow_rate(zone)
           vav_terminal = equip.to_AirTerminalSingleDuctVAVReheat.get
-          air_terminal_single_duct_vav_reheat_apply_minimum_damper_position(vav_terminal, zone_oa, has_ddc)
+          air_terminal_single_duct_vav_reheat_apply_minimum_damper_position(vav_terminal, zone, zone_oa, has_ddc)
         end
       end
     end
@@ -2292,7 +2292,7 @@ class Standard
     end
 
     # Change the min flow rate in the controller outdoor air
-    controller_oa.setMinimumOutdoorAirFlowRate(0.0)
+    # controller_oa.setMinimumOutdoorAirFlowRate(0.0)
 
     # Enable DCV in the controller mechanical ventilation
     controller_mv.setDemandControlledVentilation(true)
--- a/lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb
+++ b/lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb
@@ -1808,7 +1808,7 @@ class Standard
       if air_loop_hvac.model.version < OpenStudio::VersionString.new('3.3.0')
         controller_mv.setSystemOutdoorAirMethod('VentilationRateProcedure')
       else
-        controller_mv.setSystemOutdoorAirMethod('Standard62.1VentilationRateProcedure')
+        controller_mv.setSystemOutdoorAirMethod('Standard62.1VentilationRateProcedureWithLimit')
       end
         # Change the min flow rate in the controller outdoor air
       controller_oa.setMinimumOutdoorAirFlowRate(0.0)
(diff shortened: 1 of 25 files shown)


lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb
```diff
@@ -1873,6 +1877,21 @@ def air_loop_hvac_adjust_minimum_vav_damper_positions(air_loop_hvac)
...
+        sizing_zone = zone.sizingZone
+        sizing_zone.setDesignMinimumZoneVentilationEfficiency(0.6)
+      end
+      OpenStudio.logFree(OpenStudio::Info, 'openstudio.standards.AirLoopHVAC', "For #{air_loop_hvac.name}: the multizone outdoor air calculation method was applied, and a minimum zone ventilation efficiency of 0.6 was specified using the EnergyPlus algorithm.")
+
+      return true
+    end
+
+    # Use the manual method otherwise
+
```
lymereJ, line 1894: I don't believe that this is a code requirement, we're currently doing that in `model_system_outdoor_air_sizing_vrp_method()` in `Prototype.SizingSystem.rb`. What's the reason for moving it here and below?

asparke2: It's not a code requirement, however, the previous approach, which mainly affects ComStock or others using the older prototypes, assumed that historic design practice for the pre-1980 and 1980-2004 templates was simply to not adjust minimum damper positions, which results in higher min OA requirements for those buildings. When the change was made to put the 0.6 minimum limit into `model_system_outdoor_air_sizing_vrp_method()`, I didn't notice because ComStock was still using OS 2.9.1, so that code never got called.  I think that moving the code into standard-specific methods should result in the previous behavior:

- For pre-1980 and 1980-2004, VAV systems get Standard62.1VentilationRateProcedure with no minimum zone ventilation efficiency. This likely means higher minimum OA for those systems.
- For 2004+
    - If using OS =< 3.0.0, minimum damper positions are manually adjusted to achieve a minimum zone ventilation efficiency of 0.6
    - If using OS > 3.0.0, minimum damper positions are set to achieve a minimum zone ventilation efficiency of 0.6 using the newly-available EnergyPlus built-in algorithm.
    - I have tested these two methods on a small number of buildings and confirmed that the E+ algorithm and the manual method create nearly identical results.

lymereJ: Ok, I see, thanks for the additional information. So when we worked on that section of the code (in #1112 and #776) we missed that the DOE reference building should not be using a minimum efficiency. Knowing that I think that what you're proposing makes sense but I do see that the regression and performance tests failed. Do you mind if I have a look at why we get diffs before this gets merged? Is it urgent, I might not have time to look at it before next week.

lib/openstudio-standards/standards/Standards.AirLoopHVAC.rb
```diff
@@ -1805,8 +1805,12 @@ def air_loop_hvac_enable_multizone_vav_optimization(air_loop_hvac)
       oa_system = air_loop_hvac.airLoopHVACOutdoorAirSystem.get
       controller_oa = oa_system.getControllerOutdoorAir
       controller_mv = controller_oa.controllerMechanicalVentilation
-      controller_mv.setSystemOutdoorAirMethod('VentilationRateProcedure')
-      # Change the min flow rate in the controller outdoor air
+      if air_loop_hvac.model.version < OpenStudio::VersionString.new('3.3.0')
+        controller_mv.setSystemOutdoorAirMethod('VentilationRateProcedure')
+      else
+        controller_mv.setSystemOutdoorAirMethod('Standard62.1VentilationRateProcedure')
```
lymereJ, line 1811: I think we might want to consider using `62.1VentilationRateProcedureWithLimit` here. That's the intent in ASHRAE 90.1.

asparke2: Yeah I agree, didn't see this additional enumeration.  I will update branch.  Also, this is not urgent to merge, take your time and look through the regression tests.

mdahlhausen:
@asparke2 @lymereJ the AppG pull request is merged - can you merge in from master and resolve conflicts?

asparke2:
@lymereJ I'm going to test your suggestion, I believe that it will work and if so will implement it on a new branch and close this PR. In testing, I have confirmed an approach that uses more recent EnergyPlus capabilities to autosize minimum damper positions, which would allow us to remove the damper position adjustment code in openstudio-standards. For most models, that approach results in minimum damper positions and OA rates within 1% of the current code. I've documented that approach in a feature request issue here so we don't lose that knowledge: https://github.com/NREL/openstudio-standards/issues/1391

asparke2:
Closing this PR, choosing the simpler approach in PR https://github.com/NREL/openstudio-standards/pull/1394 instead.

PR #1374, opened 2022-09-09
Check pump power and control

weilixu:
Draft PR for review and comments.

I haven't completed the EMS side for the primary / secondary loop.

Based on existing workflow, I modified a couple functions and put them under the PRM routine for:
1. Simplified function by setting PRM as precondition for these functions
2. PRM specific parameters.

The workflow can be viewed at: https://drive.google.com/file/d/1Ql7Rg1-ItIITVdNm-vQs0v-4Ga8eN2lB/view?usp=sharing


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -1,4 +1,11 @@
 class ASHRAE901PRM < Standard
+  HOT_WATER_PUMP_POWER = 19 # W/gpm
```
dmaddoxwhite, line 2: Would it be better to set these constants down in plant_loop_apply_prm_baseline_pump_power(), since that is the only place where they are used?

weilixu: That make sense. I was trying to set them as a constant so it is easier to maintain. I will move them to a function variable.

PR #1380, opened 2022-09-15
Baseline elevator power and process loads

lymereJ:
The proposed changes add baseline elevator power and related process loads based on proposed elevator design that's inferred from the user data.


dmaddoxwhite (approved):
Looks good!

PR #1385, opened 2022-10-06
Update the PRM function arguments

weilixu:
Update the PRM method arguments.

### DIFF
--- a/lib/openstudio-standards/standards/Standards.Model.rb
+++ b/lib/openstudio-standards/standards/Standards.Model.rb
@@ -19,21 +19,18 @@ class Standard
   # @note Per 90.1, the Performance Rating Method "does NOT offer an alternative compliance path for minimum standard compliance."
   # This means you can't use this method for code compliance to get a permit.
   # @param user_model [OpenStudio::model::Model] User specified OpenStudio model
-  # @param building_type [String] the building type
   # @param climate_zone [String] the climate zone
   # @param hvac_building_type [String] the building type for baseline HVAC system determination (90.1-2016 and onward)
   # @param wwr_building_type [String] the building type for baseline WWR determination (90.1-2016 and onward)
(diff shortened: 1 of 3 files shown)


lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -19,21 +19,18 @@ def define_space_multiplier
...
   # @param debug [Boolean] If true, will report out more detailed debugging output
   # @return [Bool] returns true if successful, false if not
 
   # Method used for 90.1-2016 and onward
-  def model_create_prm_stable_baseline_building(model, building_type, climate_zone, hvac_building_type, wwr_building_type, swh_building_type, custom = nil, sizing_run_dir = Dir.pwd, run_all_orients = true, unmet_load_hours_check = true, debug = false)
-    model_create_prm_any_baseline_building(model, building_type, climate_zone, hvac_building_type, wwr_building_type, swh_building_type, true, custom, sizing_run_dir, run_all_orients, unmet_load_hours_check, debug)
+  def model_create_prm_stable_baseline_building(model, climate_zone, hvac_building_type, wwr_building_type, swh_building_type, output_dir = Dir.pwd, unmet_load_hours_check = true, debug = false)
+    model_create_prm_any_baseline_building(model, "SmallOffice", climate_zone, hvac_building_type, wwr_building_type, swh_building_type, true, false, output_dir, true, unmet_load_hours_check, debug)
```
lymereJ, line 33: Couldn't we use `nil` instead of `SmallOffice`?

weilixu: @lymereJ  Thanks for the comment.
I actually ignored one use of the building type. Line 95 shows the building type is used for the model building name.
I think setting to an empty string may make more sense this case?

mdahlhausen (approved):

PR #1389, opened 2022-10-12
Include NECB reference heat pump rules (2011 - 2017)

edwardv720:
@IainMacdonald  to review


IainMacdonald:
@edwardv720 there seems to be conflicts in the expected results. this is likely due to the update of 344 that I did earlier today. Can you pull from 344 and resolve the conflicts please?

Also simulation.rb should have the 'touch' line commented out (its only us that needs this!)

Issue #1391, opened 2022-10-14
Switch to E+ autosizing of minimum damper positions for VAV terminals


asparke2:
Currently, openstudio-standards runs it's own version of the VRP calculations to adjust the minimum damper position upward for critical zones (in `air_loop_hvac_adjust_minimum_vav_damper_positions()`), which reduced the overall minimum OA for the VAV airloops. These adjusted minimum damper positions and the adjusted minimum system OA flow rate are then hard-sized in the model. This was done because EnergyPlus could not account for a target minimum ventilation efficiency in autosizing.

As of EnergyPlus 9.4.0 (OpenStudio 3.1.0), this autosizing capability has been added. Switching to use this capability would allow us to remove a large amount of complex code in `air_loop_hvac_adjust_minimum_vav_damper_positions()`. If we ever decide to remove backward compatibility with OS versions < 3.10, we could also switch to this approach and greatly reduce the complexity of the VAV damper position logic.

In order to use the EnergyPlus autosizing approach, the following fields must be set:
1. In the `AirLoopHVAC` `Sizing:System` object, `sizing_system.setSystemOutdoorAirMethod('Standard62.1VentilationRateProcedure')`
2. In the `Sizing:Zone` for each zone on the system, `sizing_zone.setDesignMinimumZoneVentilationEfficiency(min_evz)`, where `min_evz` is the minimum ventilation efficiency used to increase minimum damper positions. This gives EnergyPlus a target to use when adjust the minimum damper position upward when autosizing. Current assumption is that for DOE-pre-1980 and 1980-2004, the dampers were NOT adjusted, implying `min_evz = 0` and for all other templates, `min_evz = 0.6`
3. In the `Sizing:Zone` for each zone on the system, `sizing_zone.setCoolingMinimumAirFlowFraction(min_damper_position)`, where `min_damper_position` is determined by `air_terminal_single_duct_vav_reheat_minimum_damper_position()`. This gives EnergyPlus a minimum flow fraction to use when autosizing the minimum damper position.
4. In the `AirTerminal:SingleDuct:VAV:Reheat` or `AirTerminal:SingleDuct:VAV:NoReheat` for each zone on the system, tell EnergyPlus to Autosize the minimu damper position using `terminal.autosizeConstantMinimumAirFlowFraction()`. This autosizing will use the info in `Sizing:Zone` object, starting with the minimum damper position as the lower bound and increasing to meet the minimum Evz as required.

[...]

PR #1394, opened 2022-10-19
Fixes VRP enumerations for newer E+ versions and makes Evz template-dependent

### LINKED ISSUE #1372: VentilationRateProcedure enums changed

asparke2:
Fixes VRP enumerations for newer E+ versions and makes Evz template-dependent. Fixes #1372 using the method suggested by @lymereJ. Opened a separate feature request issue #1391 to someday use EnergyPlus to autosize minimum damper position.

### DIFF
--- a/lib/openstudio-standards/prototypes/common/objects/Prototype.SizingSystem.rb
+++ b/lib/openstudio-standards/prototypes/common/objects/Prototype.SizingSystem.rb
@@ -57,14 +57,20 @@ class Standard
     end
 
     sizing_system = air_loop_hvac.sizingSystem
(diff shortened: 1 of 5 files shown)


asparke2:
I tested this on a large ComStock run and it appears to work as previously.

Issue #1395, opened 2022-10-20
OS SDK 3.5.0 API DOE Prototype Issues


mdahlhausen:
**ISSUE 1: zone to space translation**
**Summary:**
Recent versions of EnergyPlus added Space objects.  Some objects attached to a zone get attached to a space upon OpenStudio translation to EnergyPlus.  When this happens, it can orphan objects previously attached to specific zone object names, such as EMS controls using zone daylight sensors.  

That will generate a confusing error like this:
`Invalid Actuated Component Unique Name =CAFETERIA_ZN_1_FLR_1 ZN PRIMARYSCHOOL CAFETERIA LIGHTS  Entered in EnergyManagementSystem:Actuator=CAFETERIA_ZN_1_FLR_1ZN_LIGHT1_ACTUATOR  Component Unique key name not found   Use Output:EnergyManagementSystem object to create .edd file for valid component names.\
`
One fix is to not translate zones to spaces and instead keep them as zones.  This is enabled by setting `no_space_translation: false` in the `run_options` of an .osw.  [See example](https://github.com/NREL/OpenStudio-workflow-gem/blob/develop/spec/openstudio/workflow/workflow_analysis_ft_options_spec.rb#L60-L73).  The fix isn't durable, as it requires all downstream end users add that run option, and still leaves and issue where zone EMS objects may get duplicated to multiple Space objects if the zone has multiple spaces.

The preferred solution is to have a switch statement depending on OS version, and attach lighting sensors or related objects to Space objects directly for recent OS versions.
Code that needs to change are the occupancy controls in [ASHRAE 90.1-2016](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/ashrae_90_1/ashrae_90_1_2016/ashrae_90_1_2016.Model.rb#L261-L367) and [ASHRAE 90.1-2019](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/ashrae_90_1/ashrae_90_1_2019/ashrae_90_1_2019.Model.rb#L261-L368).

There may be other locations in the code where this is an issue, particularly around daylight sensor placement.

**ISSUE 2**: occupant diversity defaults now explicit in OS:Sizing:System
**Summary:**
Recent versions of OpenStudio now include an [Occupant Diversity](https://raw.githubusercontent.com/NREL/OpenStudio/develop/resources/model/OpenStudio.idd) field in the `OS:Sizing:System` object.
```
[...]

PR #1405, opened 2022-11-15
Adding assert to handle model generating unsuccessfully case in the r…

weilixu:
…emove_transfer function

Addressing the #1395  item number 3.
Adding assert to handle exception when the model generation is failed.


weilixu:
The test results seems PASS all tests but still the report shows 1 assertion.
@mdahlhausen Any idea what causes the CI failure?

mdahlhausen (changes requested):
There is an issue with model generation (path names I believe?) and we should solve that issue rather setting up an integration test assert that model generation fails.

We can add an error statement here so that the model generation fails gracefully.  Use an error statement, not an '`assert`'

PR #1410, opened 2023-01-30
Remove obsolete OpenStudio Model Object autosizing and hardsizing methods

asparke2:
Methods to get autosized values for component capacities, flow rates, etc. were needed by OpenStudio-Standards in order to apply code-baseline efficiencies and controls. These methods were originally implemented in Ruby inside OpenStudio-Standards for expedience and prototyping. In OpenStudio 2.5.0 (released 3/30/2018), these methods were added to the OpenStudio C++ API, rendering the Ruby implementation obsolete. However, the OpenStudio-Standards implementation was kept for backwards compatibility. Today, the expectation is that most users of OpenStudio-Standards are using a version of OpenStudio >= 2.5.0, making the Ruby implementation unnecessary.

This pull request removes the redundant Ruby methods that are present (with the same method names and signatures) in the OpenStudio C++ API. After this change, Measures or other code using `require 'openstudio-standards'` will be using the C++ methods maintained by the OpenStudio development team. This will both reduce the maintenance burden on OpenStudio-Standards and reduce confusion about the source of the problem if a method is not working properly.


lib/openstudio-standards/hvac_sizing/Siz.CoilCoolingWater.rb
```diff
@@ -2,108 +2,12 @@
...
-  end
-
-  # returns the autosized inlet air humidity ratio as an optional double
-  def autosizedDesignInletAirHumidityRatio
-
-    return self.model.getAutosizedValue(self, 'Design Size Design Inlet Air Humidity Ratio', '')
-    
-  end
+  # returns the autosized design coil load
+  def autosizedDesignCoilLoad
```
mdahlhausen, line 6: why does this method remain?

lib/openstudio-standards/standards/necb/NECB2011/hvac_systems.rb
```diff
@@ -42,7 +42,7 @@ def air_loop_hvac_economizer_required?(air_loop_hvac)
 
     # puts air_loop_hvac.name.to_s
     # Design Supply Air Flow Rate: This method below reads the value from the sql file.
-    dsafr_m3_per_s = air_loop_hvac.model.getAutosizedValue(air_loop_hvac, 'Design Supply Air Flow Rate', 'm3/s')
+    dsafr_m3_per_s = air_loop_hvac.getAutosizedValue('Design Supply Air Flow Rate', 'm3/s')
```
mdahlhausen, line 45: why not `air_loop_hvac.autosizeDesignSupplyAirFlowRate`?

lib/openstudio-standards/standards/necb/NECB2015/hvac_systems.rb
```diff
@@ -125,18 +125,18 @@ def apply_maximum_loop_pump_power(model)
             max_powertoload = model_find_object(max_total_loop_pump_power_table, search_hash)['total_normalized_pump_power_wperkw']
           when 'OS_Pump_VariableSpeed'
             pumps << supplycomp.to_PumpVariableSpeed.get
-            total_pump_power += model.getAutosizedValue(supplycomp, 'Design Power Consumption', 'W').to_f
+            total_pump_power += supplycomp.getAutosizedValue('Design Power Consumption', 'W').to_f
```
mdahlhausen, line 128: why not `supplycomp.ratedPowerConsumption`?

test/necb/unit_tests/tests/test_necb_pumppower_rules.rb
```diff
@@ -96,16 +96,16 @@ def test_NECB2015_pumppower
               max_powertoload = 22
             when 'OS_Pump_VariableSpeed'
               pumps << supplycomp.to_PumpVariableSpeed.get
-              total_pump_power += model.getAutosizedValue(supplycomp, 'Design Power Consumption', 'W').to_f
+              total_pump_power += supplycomp.getAutosizedValue('Design Power Consumption', 'W').to_f
```
mdahlhausen, line 99: why not `supplycomp.ratedPowerConsumption`?

mdahlhausen (commented):
Should any of the methods in these files be moved to OpenStudio?:

- Siz.AirLoopHBAC.rb
- Siz.CoilCoolingWater.rb
- Siz.HeatingCoolingFuels.rb
- Siz.HVACComponent.rb
- Siz.Model.rb
- Siz.ThermalZone.rb

If not, can they be moved to elsewhere in standards? (e.g. heating and cooling fuels make sense in utilities)

Any reason why the necb/NRCAN methods weren't updated to direct methods in OpenStudio?

lib/openstudio-standards/standards/necb/NECB2015/hvac_systems.rb
asparke2: good call, changed

test/necb/unit_tests/tests/test_necb_pumppower_rules.rb
asparke2: good call, changed

asparke2:
Not using the component-specific methods that exist was an oversight, those have been corrected.

I made [a feature request in OpenStudio](https://github.com/NREL/OpenStudio/issues/4792) to implement the other outstanding methods in C++, but in the meantime will keep the implementation as extensions to OpenStudio::Model objects inside /hvac_sizing so that when the C++ implementation happens, the methods are easy to find and swap out.

- Siz.AirLoopHBAC.rb, Siz.CoilCoolingWater.rb, Siz.ThermalZone.rb
  - The methods in these files probably should be implemented in C++. They weren't, either because they are methods that don't correspond to input fields found in the objects, or because they come from a sizing-related object such as `Sizing:Zone` that controls another object such as `ThermalZone`.

- Siz.HeatingCoolingFuels.rb, Siz.HVACComponent.rb
  - These methods could be implemented in C++, but the developers will need to keep a close eye when new objects are added because these methods can be difficult to test.

- Siz.Model.rb
  - Simply loads the other files. Will be unnecessary when methods are in C++.

mdahlhausen:
@asparke2 looks like the latest commit introduced an error in the NECB workflow

Issue #557, opened 2018-12-07
Heat Pump Water Heater Support


mdahlhausen:
**Summary**
Add wrapped-condenser and pumped-condenser heat pump water heater object methods to openstudio-standards.

**Background**
Heat pump water heaters (HPWHs) are a key technology for grid decarbonization.  They are currently modeled in openstudio-standards by adjusting the part load efficiency curve for an electric mixed water heater object.  This is substantially faster than using a stratified tank method that the residential teams uses to model small wrapped-condenser HPWHs.  Small commercial establishments may use a residential-type wrapped-condenser HPWH, while larger establishments use a pumped-condenser HPWH placed outdoors.

**Tasks**
 - [ ] Decide whether to keep the part load efficiency curve adjustment method, or use the stratified tank model
 - [ ] Determine standard efficiencies to use for commercial type wrapped-condenser and pumped-condenser HPWHs
 - [ ] Determine prevalence of wrapped-condenser and pumped-condenser variants
 - [ ] Project future efficiency levels for HPWHs
 
If using the stratified tank method:
 - [ ] update sizing method in openstudio-standards to scale HPWH (current hard-sized for single 80 gal tank)
 - [ ] add separate methods for wrapped-condenser and pumped-condenser variants
 - [ ] research performance curves for pumped condensers

**Notes from 18 Oct 2018 meeting with residential team**
- the residential team is considering speed-up methods for wrapped condenser modeling in EnergyPlus, perhaps using a TRNSYS modeling, with a goal to finish this work in 6-12 months ahead of the next major E+ release.  They may use our approach of the part load efficiency curve.
- the full residential heat pump water heater measure allows the user to account for ducted units, in the case the water heater condenser exchanges air in a zone it is not physically located in
- a separate task is determine how to model losses from distribution systems for central HPWHs.  BEOpt has some methods to handle re-circulation loops.
- performance rating methods for heat pump water heaters use UEF, which is similar to the old energy factor (EF) method, but uses a different draw profile
- 2nd generation heat pump water heaters are much more efficient than 1st generation
- Ben Larson and Jeff Robertson are the main points of contact

mdahlhausen:
Added with #1420.
Additional work would be to add different specifications for HPWHs of different sizes, and breakout larger sizes into a bank of HPWHs.

PR #1420, opened 2023-02-21
adding hpwh ems schedule override

JanghyunJK:
**Summary**
- Coming from EUSS hpwh measure development
- Measure dev work details are [here](https://nrel.sharepoint.com/:w:/r/sites/comstock/Shared%20Documents/ComStock%20Filing%20Cabinet/Efforts/Measures/SWH%20-%20HPWH/SWH%20-%20HPWH_v5.docx?d=w16ff1868cc694ce4b6a964e339cf087e&csf=1&web=1&e=LOIAoM) (for ComStock team only)
- And this PR is mostly to fill the gap of adding EMS for overriding setpoint schedules of HPWH
- Additions in this PR is a slightly modified version of [this](https://github.com/NREL/OpenStudio-HPXML/blob/7e9429ac8f288a22aabbd1d4b91aa0b781564522/HPXMLtoOpenStudio/resources/waterheater.rb#L949-L1018) from HPXML repo
- Changes are not tested in terms of code execution..

**Next Steps** (mostly to @mdahlhausen)
- This is the first time (other than updating jsons) for me to actually make code changes so really unsure how to test these changes. So would be great to know if there's a way (e.g., like a unit test for an os measure) to properly test the changes.
- there's a [part](https://github.com/NREL/openstudio-standards/commit/cee6f6caaf590466d733c84b59bc7f2841f2215c) where I added a simple string conversion so that a string can be converted to EMS-friendly string. Let me know if there's any existing method for this.


JanghyunJK:
hm sounds like I'd also have to make some changes on the test ruby code?

mdahlhausen (changes requested):
Thanks for making these changes @JanghyunJK!  I've been wanting to get better HPWH operation for the model_add_heatpump_water_heater method for a while.

I made some pretty heavy edits, and it still needs some testing:
- the ems code only applies to stratified tank models with the wrapped condenser, so I added some logic to check whether it applies
- added a user argument for if they want to use the ems code, defaulted to false
- moved the ems_friendly_name to the utilities section and made some edits to it. There isn't a method like it yet, but there a lot of methods in standards that use EMS that could benefit from the method. I'll make a separate issue and fix for all of those changes
- removed the methods calls and placed the content in line. I generally don't like making methods if they are only used one place in code; it makes it harder to follow. I do make some exceptions for very long methods, but didn't feel this one needed it.
- fixed the EMS syntax

This still needs some testing. I'd like to apply this to a model and run it as an integration test in case any of the objects change in the future. I'll add that next before merging.

mdahlhausen:
related to #557

PR #1428, opened 2023-03-08
Appendix g dev debug unmet load hour issues

weilixu:
Bug fixes including
1. Add increment values that breaks the unmet load hour break loop
2. Add statement to close SQL sessions before running a new simulation in the same folder (issues under Window OS)
3. Add error handler when simulation failed in the unmet load hour loop.
4. Add schedule type check for optimal start EMS code

### DIFF
--- a/lib/openstudio-standards/standards/Standards.Model.rb
+++ b/lib/openstudio-standards/standards/Standards.Model.rb
@@ -454,17 +454,18 @@ class Standard
       # Check unmet load hours
       if unmet_load_hours_check
         nb_adjustments = 0
-        loop do
-          # Loop break condition: Limit the number of zone sizing factor adjustment to 8
+        loop d
+          # Loop break condition: Limit the number of zone sizing factor adjustment to 3
           unless nb_adjustments < 3
-            OpenStudio.logFree(OpenStudio::Error, 'openstudio.standards.Model', "After 8 rounds of zone sizing factor adjustments the unmet load hours for the baseline model (#{degs} degree of rotation) still exceed 300 hours. Please open an issue on GitHub (https://github.com/NREL/openstudio-standards/issues) and share your user model with the developers.")
+            OpenStudio.logFree(OpenStudio::Error, 'openstudio.standards.Model', "After 3 rounds of zone sizing factor adjustments the unmet load hours for the baseline model (#{degs} degree of rotation) still exceed 300 hours. Please open an issue on GitHub (https://github.com/NREL/openstudio-standards/issues) and share your user model with the developers.")
             break
           end
-          # Close the previous SQL session if any - this could prevent EnergyPlus from overloading the same session
+          # Close the previous SQL session if open to prevent EnergyPlus from overloading the same session
           sql = model.sqlFile.get
           if sql.connectionOpen
             sql.close
           end
+
           if model_run_simulation_and_log_errors(model, "#{sizing_run_dir}/final#{degs}")
             # If UMLH are greater than the threshold allowed by Appendix G,
             # increase zone air flow and load as per the recommendation in
--- a/lib/openstudio-standards/standards/Standards.Model.rb
+++ b/lib/openstudio-standards/standards/Standards.Model.rb
@@ -454,7 +454,7 @@ class Standard
       # Check unmet load hours
       if unmet_load_hours_check
         nb_adjustments = 0
-        loop d
+        loop do
           # Loop break condition: Limit the number of zone sizing factor adjustment to 3
           unless nb_adjustments < 3
             OpenStudio.logFree(OpenStudio::Error, 'openstudio.standards.Model', "After 3 rounds of zone sizing factor adjustments the unmet load hours for the baseline model (#{degs} degree of rotation) still exceed 300 hours. Please open an issue on GitHub (https://github.com/NREL/openstudio-standards/issues) and share your user model with the developers.")
--- a/lib/openstudio-standards/standards/Standards.Model.rb
+++ b/lib/openstudio-standards/standards/Standards.Model.rb
@@ -455,66 +455,74 @@ class Standard
       if unmet_load_hours_check
         nb_adjustments = 0
         loop do
-          model_run_simulation_and_log_errors(model, "#{sizing_run_dir}/final#{degs}") == false
-          # If UMLH are greater than the threshold allowed by Appendix G,
-          # increase zone air flow and load as per the recommendation in
-          # the PRM-RM; Note that the PRM-RM only suggest to increase
-          # air zone air flow, but the zone sizing factor in EnergyPlus
-          # increase both air flow and load.
-          if model_get_unmet_load_hours(model) > 300
-            # Limit the number of zone sizing factor adjustment to 8
-            unless nb_adjustments < 8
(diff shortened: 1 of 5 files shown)


lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -455,66 +455,74 @@ def model_create_prm_any_baseline_building(user_model, building_type, climate_zo
...
+            OpenStudio.logFree(OpenStudio::Error, 'openstudio.standards.Model', "After 8 rounds of zone sizing factor adjustments the unmet load hours for the baseline model (#{degs} degree of rotation) still exceed 300 hours. Please open an issue on GitHub (https://github.com/NREL/openstudio-standards/issues) and share your user model with the developers.")
+            break
+          end
+          # Close the previous SQL session if any - this could prevent EnergyPlus overloading the same session
+          sql = model.sqlFile.get
+          if sql.connectionOpen
+            sql.close
+          end
+          if model_run_simulation_and_log_errors(model, "#{sizing_run_dir}/final#{degs}")
+            # model_run_simulation_and_log_errors(model, "#{sizing_run_dir}/final#{degs}") == false
```
lymereJ, line 469: Maybe we should remove that line.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -455,66 +455,74 @@ def model_create_prm_any_baseline_building(user_model, building_type, climate_zo
...
+                  # Make adjustment to zone heating sizing factor
+                  # Do not adjust factors greater or equal to 2
+                  # Rewrite the logic to avoid 5 level nested logic.
+                  htg_umlh > 150 ? sizing_factor = [2.0, sizing_factor * 1.1].min : sizing_factor = [2.0, sizing_factor * 1.05].min
+                  # if sizing_factor < 2.0
+                  #  if htg_umlh > 150
+                  #    sizing_factor *= 1.1
+                  #  else #elsif htg_umlh > 50 is not needed for the same reason
+                  #    sizing_factor *= 1.05
+                  #  end
```
lymereJ, line 515: Maybe we should remove these comments?

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -455,66 +455,74 @@ def model_create_prm_any_baseline_building(user_model, building_type, climate_zo
...
                   end
+                  # Make adjustment to zone cooling sizing factor
+                  # Do not adjust factors greater or equal to 2
+                  # Rewrite the logic to avoid 5 level nested logic.
+                  clg_umlh > 150 ? sizing_factor = [2.0, sizing_factor * 1.1].min : sizing_factor = [2.0, sizing_factor * 1.05].min
+                  # if clg_umlh > 150
+                  #  sizing_factor *= 1.1
+                  # else # elseif clg_umlh > 50 is not needed since the code block only reached when clg_umlh>50
+                  #  sizing_factor *= 1.05
+                  # end
```
lymereJ, line 492: Maybe we should remove these comments?

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -455,66 +455,74 @@ def model_create_prm_any_baseline_building(user_model, building_type, climate_zo
...
-                  if clg_umlh > 150
-                    sizing_factor *= 1.1
-                  elsif clg_umlh > 50
-                    sizing_factor *= 1.05
+                  if thermal_zone.sizingZone.zoneCoolingSizingFactor.is_initialized
+                    sizing_factor = thermal_zone.sizingZone.zoneCoolingSizingFactor.get
                   end
+                  # Make adjustment to zone cooling sizing factor
+                  # Do not adjust factors greater or equal to 2
+                  # Rewrite the logic to avoid 5 level nested logic.
```
dmaddoxwhite, line 486: Don't need to keep "Rewrite... " comment.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -455,66 +455,74 @@ def model_create_prm_any_baseline_building(user_model, building_type, climate_zo
...
-                  elsif htg_umlh > 50
-                    sizing_factor *= 1.05
+                  if thermal_zone.sizingZone.zoneHeatingSizingFactor.is_initialized
+                    # Get zone heating sizing factor
+                    sizing_factor = thermal_zone.sizingZone.zoneHeatingSizingFactor.get
                   end
+
+                  # Make adjustment to zone heating sizing factor
+                  # Do not adjust factors greater or equal to 2
+                  # Rewrite the logic to avoid 5 level nested logic.
```
dmaddoxwhite, line 508: Remove "Rewrite ... " comment.

dmaddoxwhite (approved):

mdahlhausen:
@weilixu @lymereJ 8 is a lot of adjustments; that's a lot of time a user will be at their PC while the measure is looping through sizing runs.  I have a commit drafted to set it to three instead.  Is that ok?

lymereJ:
> @weilixu @lymereJ 8 is a lot of adjustments; that's a lot of time a user will be at their PC while the measure is looping through sizing runs. I have a commit drafted to set it to three instead. Is that ok?

@dmaddoxwhite, is that okay?

weilixu:
> s; that's a lot of time a user will be at their PC while the measure is looping through sizing runs. I have a commit drafted to set it to three instead. Is that ok?

I am ok with it, I wonder if we should bump up the multiplier to account for the reduced iterations.

lymereJ:
The multipliers are from the PRM-RM. So we probably would want to specify that we changed the approach if that's what we want to do.

dmaddoxwhite:
How about adding an argument to the method to set the maximum number of sizing factor adjustment runs? If we limit to three, without an override, a user might get a model that fails due to unmet hours with no way to fix it.

mdahlhausen:
@dmaddoxwhite in my experience, equipment size is rarely if ever the source of unmet hours.  If one iteration doesn't fix it, eight iterations won't either.  The issue is more likely to do with schedule alignment (e.g. 5% occupancy when the HVAC system is turned off), design conditions (extremes in the annual weather file very different from design conditions), or some weird VAV sizing issue.  For many of these, even an HVAC systems with 10x the capacity isn't going to solve them.

Users will generate models that have baselines that fail the UMLH test.  I'd rather they fail faster (3 iterations) than wait for 8 and wonder why it still didn't work.  Alternatively, we can add some QAQC checks to point out basic issues (like 5% occupancy in zones when the HVAC system is off).  I've already got methods in standards to check for those sorts of things.

mdahlhausen (approved):
looks fine; made some minor cleanup fixes and changed number of adjustments to 3.  It may be worth adding some checking in here to pick out the cause of unmet hours, rather than just increasing sizing factors.  The PRM RM includes some direction beyond sizing, but IMO it's not very specific or actionable.

mdahlhausen:
@weilixu @lymereJ @dmaddoxwhite good to merge.  Up to you if you want to edit the # of adjustments further or make that a user argument.  It's also fine as is.

weilixu:
> @weilixu @lymereJ @dmaddoxwhite good to merge. Up to you if you want to edit the # of adjustments further or make that a user argument. It's also fine as is.

Thanks @mdahlhausen, we will update the number of iteration.
We also need to disable an optimal control in the PRM method. Will make a push shortly.

PR #1430, opened 2023-03-09
Update simulation.rb

mdahlhausen:
close the sql file if open in addition to resetting the sql file

### DIFF
--- a/lib/openstudio-standards/utilities/simulation.rb
+++ b/lib/openstudio-standards/utilities/simulation.rb
@@ -36,10 +36,6 @@ Standard.class_eval do
     epw_path = epw_path.get
 
     # close current sql file
-    if model.sqlFile.is_initialized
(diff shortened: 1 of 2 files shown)


weilixu:
The test_infiltration function failed. I am investigating the error in prm test file

weilixu:
The SQL issue is fixed, 3.2.1 pass CI test @mdahlhausen

mdahlhausen:
@weilixu I'm doing some more checks before I merge this.  I'm worried it could cause some issues in other projects that use standards, similar to the initial failure we saw on the PRM tests on the CI.  I may just remove the changes to `simulation.rb` and keep the changes you made to `test_appendix_g_prm.rb`.  Is that ok?

mdahlhausen:
May reopen this and reapply the changes to `simulation.rb` in the future after more testing.

PR #1434, opened 2023-03-10
Incorporating NRCan Changes

ckirney:
The major changes in this are:
-Adding support for NECB 2020
-Move to OpenStudio 3.5.1
-Addition of heat pumps to NECB reference buildings (rather than just as an ECM)
-Removal and addition to regression and unit tests


mdahlhausen (changes requested):
Diff is to large to review the files, and a bundler gem spec change is making this fail on the CI.
1)	Revert the changes to the bundler spec so that it can run on the CI
2)	Split out some commits into a separate pull request. My suggestion is find a commit or two where you delete or alter many files, e.g. https://github.com/NREL/openstudio-standards/pull/1434/commits/a79450669dd4825e760b12a59b2c4162151371c0, and make those a separate pull request.  Also, please delete old/extraneous code rather than just commenting it out.

PR #1439, opened 2023-03-14
Merge from commit eea3c93

ckirney:


mdahlhausen (approved):
Looks great @ckirney.  I will review `apply_kiva_foundation(model)` in more detail as I dig into the differences between OSv3.2.1 and OSv3.5.1.

PR #1447, opened 2023-03-14
break out lights in EMS code

### LINKED ISSUE #1446: 90.1-2016 and 90.1-2019 lighting EMS code
This is issue is already discussed somewhat as issue 1 in #1395.

In OpenStudio v3.3.0 and above, most 90.1-2016 and 90.1-2019 models will fail at runtime with an error like this:
`Invalid Actuated Component Unique Name =CAFETERIA_ZN_1_FLR_1 ZN PRIMARYSCHOOL CAFETERIA LIGHTS  Entered in EnergyManagementSystem:Actuator=CAFETERIA_ZN_1_FLR_1ZN_LIGHT1_ACTUATOR  Component Unique key name not found   Use Output:EnergyManagementSystem object to create .edd file for valid component names.\

It's from the way EMS actuators are translated in the forward translator from OS to E+.  The relevant code is [here](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/ashrae_90_1/ashrae_90_1_2016/ashrae_90_1_2016.Model.rb#L261-L367) and [here](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/prototypes/ashrae_90_1/ashrae_90_1_2019/ashrae_90_1_2019.Model.rb#L261-L368).

Initially, this appeared as an [OpenStudio issue](https://github.com/NREL/OpenStudio/issues/4786).  But as I describe in [this comment](https://github.com/NREL/OpenStudio/issues/4786#issuecomment-1468905643), EnergyPlus is actually right in it's failure, as the code in early versions were actuating the same component.

The issue is that the lights object is defined at the SpaceType level and shared among many spaces / zones.  The EMS control needs to operate on one unique lights object per zone/space.

@lymereJ @weilixu I'll take a shot at addressing this and @ you when I've got it refactored within the next day or so.

mdahlhausen:
- clone and assign individual lights objects to each space
- use the space object in E+9.6 / OS v3.3 and higher versions
- hardcode space floor area as a number instead of an internal variable (will need to delete and re-run method if space sizes change)
Fixes #1446 and issue 1 listed in #1395.

### DIFF
--- a/lib/openstudio-standards/prototypes/ashrae_90_1/ashrae_90_1_2016/ashrae_90_1_2016.Model.rb
+++ b/lib/openstudio-standards/prototypes/ashrae_90_1/ashrae_90_1_2016/ashrae_90_1_2016.Model.rb
@@ -286,34 +286,53 @@ class ASHRAE9012016 < ASHRAE901
       end
       space = spaces[0]
       space_lights = space.lights
+      lights_defined_by_spacetype = false
       if space_lights.empty?
         space_lights = space.spaceType.get.lights
+        lights_defined_by_spacetype = true
       end
       space_people = space.people
       if space_people.empty?
         space_people = space.spaceType.get.people
       end
 
-      next if space_lights.empty? # skip space with no lights
+      # guard clause to skip space with no lights
+      next if space_lights.empty? 
+
+      # if lights are defined at the space type level, clone each lights object and make it individual to the space
+      new_space_lights = []
+      if lights_defined_by_spacetype
+        space_lights.each do |lights|
+          new_lights = lights.clone.to_Lights.get
+          new_lights.setName("#{space.name}-#{lights.name}")
+          new_lights.setSpace(space)
+          new_space_lights << new_lights
+        end
+        space_lights = new_space_lights
+      end
 
       zone_name = zone.name.to_s
       next if zone_name =~ /data\s*center/i # skip data centers
 
-      # EnergyPlus v9.4 name change for EMS actuators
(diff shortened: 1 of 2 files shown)


lymereJ:
I'll let @leijerry888 review it since he's the author of that method.

mdahlhausen:
Known error in the secondary school prototype in OSv3.2.1:
The tests pass on a windows machine, and fail on a linux machine.  Looking into it, but merging anyways b/c it is likely an OpenStudio bug that needs to be addressed, not an openstudio-standards issue.

```
  test_regression-SecondarySchool-90.1-2016-ASHRAE_169-2013-4A    FAIL (36.48s)
        For OS:Daylighting:Control called 'Aux_Gym_ZN_1_FLR_1 Daylt Sensor 2'  'Position X-Coordinate': true model = '12.0', compare model = '3.0'
        For OS:Daylighting:Control called 'Aux_Gym_ZN_1_FLR_1 Daylt Sensor 2'  'Position Y-Coordinate': true model = '8.65277', compare model = '2.8889'
        For OS:ThermalZone called 'Aux_Gym_ZN_1_FLR_1 ZN'  'Fraction of Zone Controlled by Primary Daylighting Control': true model = '0.999', compare model = '0.5'
        For OS:ThermalZone called 'Aux_Gym_ZN_1_FLR_1 ZN'  'Fraction of Zone Controlled by Secondary Daylighting Control': true model = '0.0', compare model = '0.499'
```

PR #1458, opened 2023-03-15
NRCan merge using OS351

ckirney:
This commit is our latest and all NRCan tests should work.  This also reverts the bundler version to 2.1.4.


mdahlhausen (approved):
Only NECB tests are failing, because the regression model version (OSv3.5.1) is newer than the CI at OvS3.2.1.
I'll work through the few merge conflicts and merge this.

mdahlhausen:
Merge conflicts were only in btap.model.rb.  I'll run it through the CI one more time just to be sure, but I think this is good to merge!

mdahlhausen:
This pull request will knowingly cause the both the OSv3.2.1 and OSv3.5.1 CI to show failures.  There should be no NRCAN failures on OSv3.5.1 CI and no non-NRCAN failures on the OSv3.2.1 CI.

PR #1471, opened 2023-03-24
Logging and test update

lymereJ:
Use logging instead or `print()` and expand a bit on the tests to make sure that JSON generated from a database created using either CSV of JSON files has the same content.

### DIFF
--- a/data/openstudio_standards_data/test/test_database.py
+++ b/data/openstudio_standards_data/test/test_database.py
@@ -2,12 +2,18 @@ from unittest import TestCase
 from unittest import mock
 import os
 import sqlite3
+import difflib
+import glob
+import shutil
 from applications.database_maintenance import (
     create_openstudio_standards_database_from_csv,
+    create_openstudio_standards_database_from_json,
+    export_openstudio_standards_database_to_csv,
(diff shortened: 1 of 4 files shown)


data/openstudio_standards_data/test/test_database.py
```diff
@@ -44,5 +60,47 @@ def test_create_database():
...
-    # Close DB
+    # Create a copy of original JSON files
+    if os.path.isdir("./original_database_files"):
+        shutil.rmtree("./original_database_files", ignore_errors=True)
+    shutil.copytree("./database_files", "./original_database_files")
+
+    # Export data to JSON and CSV files
+    export_openstudio_standards_database_to_json(conn, save_dir="./database_files/")
+    export_openstudio_standards_database_to_csv(conn, save_dir="./database_files/")
     conn.close()
```
weilixu, line 71: Do we need to close the connection here? It seems the conn is used in line 84 and 87

lymereJ: the `conn` on line 71 is different than the one on lines 85 and 88.

PR #1476, opened 2023-03-27
Appx debug

weilixu:
Large office is still failing in the hvac test due to a known issue with Eplus 22.2.

Shall we temporarily remove this test case from the unit test?


mdahlhausen:
@weilixu, @lymereJ pointed out the large office failures are because of [this](https://github.com/NREL/EnergyPlus/pull/9806) issue, and I don't see the test failures in OSv3.6.0 which uses E+ v23 with the fix.

There is a larger problem of how to handle setting system capacities and efficiencies when the design load and airflow is zero.  It's a source of errors in other models too.

One option is to set a minimum size for each kind equipment.  It could be a universal size minimum, say 1/2 ton, ~1.5 kW, which is about the smallest equipment size one can find for space heaters or residential a/c units.

A second option is to remove equipment entirely if there is no load.  I prefer this option, but it may mean a lot of tricky clean up for model objects if the user selects complex HVAC systems with EMS controls for small zones with no load.

mdahlhausen (approved):

PR #1481, opened 2023-03-30
Remove Large Office in PRM unit tests

weilixu:


mdahlhausen:
@weilixu The LargeOffice tests are passing on develop (OS v3.6.0).  I'm ok keeping them in, maybe adding specific logic to not test them with OS v3.5.1 or earlier versions instead of removing or disabling them.

mdahlhausen (changes requested):
@weilixu The LargeOffice tests are passing on develop (OS v3.6.0). I'm ok keeping them in, maybe adding specific logic to not test them with OS v3.5.1 or earlier versions instead of removing or disabling them.

weilixu:
> @weilixu The LargeOffice tests are passing on develop (OS v3.6.0). I'm ok keeping them in, maybe adding specific logic to not test them with OS v3.5.1 or earlier versions instead of removing or disabling them.

The purpose of removing the LargeOffice is that the failure could cause the test unit miss some other test cases.
Adding OS v3.5.1 version check make sense. I can update that so it become a more permanent solution. Thanks!

PR #1500, opened 2023-04-24
Fix/issue1440

mdahlhausen:


mdahlhausen:
update regression models once OSv3.6 is released

mdahlhausen:
Re-do branch with from updated regression tests and radiant controls updates.

PR #1501, opened 2023-04-25
check PTHP cop coefficients

mdahlhausen:
90.1-2016 and 90.1-2019 PTHP systems don't use the pthp_cop_coefficient_1, pthp_cop_coefficient_2 coefficients as earlier versions.  Add a check that they are present for earlier standards.

### DIFF
--- a/lib/openstudio-standards/standards/Standards.CoilHeatingDXSingleSpeed.rb
+++ b/lib/openstudio-standards/standards/Standards.CoilHeatingDXSingleSpeed.rb
@@ -129,7 +129,7 @@ class Standard
     end
 
     # If PTHP, use equations
-    if sub_category == 'PTHP'
+    if sub_category == 'PTHP' && !ac_props['pthp_cop_coefficient_1'].nil? && !ac_props['pthp_cop_coefficient_2'].nil
       pthp_cop_coeff_1 = ac_props['pthp_cop_coefficient_1']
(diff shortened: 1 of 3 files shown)


lib/openstudio-standards/standards/Standards.CoilHeatingDXSingleSpeed.rb
```diff
@@ -129,7 +129,7 @@ def coil_heating_dx_single_speed_standard_minimum_cop(coil_heating_dx_single_spe
     end
 
     # If PTHP, use equations
-    if sub_category == 'PTHP'
+    if sub_category == 'PTHP' && !ac_props['pthp_cop_coefficient_1'].nil? && !ac_props['pthp_cop_coefficient_2'].nil
```
lymereJ, line 132: 90.1-2019 and 90.1-2016 still use the coefficients, just not for units less than 7 kBtu/h and for the larger ones that have a capacity greater than 15 kBtu/h. An EER should be used in these cases:

image

image

Will this change prevent the EER values to be assigned if the `sub_category` is `PTHP`?

The same thing should be happening for PTAC.

mdahlhausen: All this method does is ensure that the PTHP coefficients are valid before applying them.  If other efficiency values are defined, those get applied instead a few lines later on in the method.

PR #1504, opened 2023-05-03
clean the duplicate script for boilder and chiller

yunyangye:


lib/openstudio-standards/standards/Standards.ChillerElectricEIR.rb
```diff
@@ -95,7 +95,7 @@ def chiller_electric_eir_standard_minimum_full_load_efficiency(chiller_electric_
       cop = kw_per_ton_to_cop(kw_per_ton)
     end
 
-    return cop
```
weilixu, line 98: The reason I don't really favor for this approach is that the function has three returns, and other two return statement returns nil.
As the function is called in the `chiller_electric_eir_apply_efficiency_and_curves`
```ruby
    cop, capacity_tons, kw_per_ton = chiller_electric_eir_standard_minimum_full_load_efficiency(chiller_electric_eir)
```
In the case of nil, wouldn't it cause more confusion when the above line tries to unpackage the return value? I am not even sure how Ruby is going to handle this error.

My suggestion would be replace the two returns with exception handler or assign default values so we are sure what is going on inside this function.

mdahlhausen: Agreed; the extra methods are only returning values for logging messages anyways. There is no need for this to be a separate method. Add logging within the method if you like.

PR #1507, opened 2023-05-09
Wx bug fix fenestration expansion missing construction name

weilixu:
We discover one issue with a sample model below (change the txt to osm to test)
[sample_med_os_ad1.txt](https://github.com/NREL/openstudio-standards/files/11426666/sample_med_os_ad1.txt)

The model does not have defaultConstructionSet assigned to space, spaceType, building or even model. However, it does run successfully.
This causes our code failure when searching for the [defaultConstructionSet ](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlanarSurface.rb#L33-L58)to fill up empty construction.

The failure cause EnergyPlus simulation failure due to missing construction name.

I suggest to handle this as assertion instead of return empty hashmap - give enough information for user to debug their model.

### DIFF
--- a/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlanarSurface.rb
+++ b/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlanarSurface.rb
@@ -54,10 +54,10 @@ class ASHRAE901PRM < Standard
         cons_set = space.model.building.get.defaultConstructionSet.get
         construction = get_default_surface_cons_from_surface_type(surface_category, surface_type, cons_set)
       end
-      OpenStudio.logFree(OpenStudio::Debug, 'prm.log', "Search for #{planar_surface.name.get} defaultConstructionSet, find #{construction ? construction.name.get: "Nil"}")
-      raise "No defaultConstructionSet defined for space #{space.name.get}, failed to find default construction for surface #{planar_surface.name.get}" unless construction
-
-      return previous_construction_map if construction.nil?
+      prm_raise(construction,
+                @sizing_run_dir,
+                "Surface #{planar_surface.name.get} does not have a construction. Failed to find defaultConstructionSet for #{planar_surface.name.get}. Add a construction for the surface or add a defaultConstructionSet to Space #{space.name.get} or SpaceType #{space_type.name.get}.",
+                "Failed to find defaultConstructionSet for #{planar_surface.name.get}. Check inputs.")
     end
 
     # Determine if residential or nonresidential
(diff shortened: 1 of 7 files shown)


dmaddoxwhite (approved):

weilixu:
I took another approach to address this issue.
I created a new assertion method to handle such issue in the PRM routine.
By calling this assert function, the OSSTD will raise PRMError exception with message.

In the measure, we can capture PRMError exception and display message to users.

In addition, the assertion function exports the prm log for debug purpose.

In order to have the log correctly exported to the sizing run directory, I added a class variable `@sizing_run_dir` in the ASHRAE901PRM class. This class variable will be set in the `handle_user_input_data`, which implemented in the `ashrae_90_1_prm.Model.rb`

Let me know if that is a desired approach and whether the implementation could be improved.

lib/openstudio-standards/utilities/assertion.rb
```diff
@@ -0,0 +1,17 @@
+# PRM generation custom Error
+class PRMError < StandardError
+end
+
+# Finds capacity in W
```
lymereJ, line 5: Is that a place holder? The description doesn't seem to match with what's in the function.

weilixu: It was a copy paste error - thanks for pointing it out!

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlanarSurface.rb
```diff
@@ -54,8 +54,10 @@ def planar_surface_apply_standard_construction(planar_surface, climate_zone, pre
         cons_set = space.model.building.get.defaultConstructionSet.get
         construction = get_default_surface_cons_from_surface_type(surface_category, surface_type, cons_set)
       end
-
-      return previous_construction_map if construction.nil?
+      prm_raise(construction,
```
dmaddoxwhite, line 57: Do we need to add the conditional to only call prm_raise if construction.nil? is true?

weilixu: @dmaddoxwhite  that is handled inside the `prm_raise` function
See: https://github.com/NREL/openstudio-standards/blob/0fe207df7d68d3a4140232320a7a275d2ae452f8/lib/openstudio-standards/utilities/assertion.rb#L12

dmaddoxwhite: I see, so if construction is nil, then it is interpreted as "FALSE" by prm_raise.

weilixu: @dmaddoxwhite That is correct - in ruby, False and nil is Falsey.
Alternatively, you can also do:
```ruby
prm_raise(construction == nil, @sizing_run_dir, ...)
```
This would explicitly expose the condition logic - I am OK with both way in this specific case.

weilixu:
All passed but the test performance tests are failed.

Issue #1530, opened 2023-06-27
Remove duplicate schedule.json files


mdahlhausen:
ComStock is duplicating the ashrae and deer schedules.json  for each comstock template.
This makes the installed gem size ~220 MB larger than needed:
![installed_size](https://github.com/NREL/openstudio-standards/assets/2249271/9e2987c2-935d-4340-a9c0-2b5de3a0636e)

mdahlhausen:
Fixed with #1758

PR #1531, opened 2023-06-27
delete duplicate comstock schedules

mdahlhausen:
also edit script to prevent them from being written again


mdahlhausen:
meant to fix #1530 

The current implementation removes all comstock schedules, and defaults to the deer/ashrae schedules. We need some different higher level architecture to retain the unique comstock schedules, but not duplicate them for each template.

mdahlhausen:
Leaving in until a refactor to significantly rework how ComStock schedules are stored/applied.

PR #1549, opened 2023-07-26
Appx g/ehancement/update ground temperature profile

weilixu:


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -2087,32 +2086,35 @@ def model_update_ground_temperature_profile(model, climate_zone)
       end
     end
 
-    return false unless surfaces_with_fc_factor_boundary
```
lymereJ, line 2090: Since we change the return value we should probably edit the YARD comment for the method.

lymereJ (approved):

PR #1486, opened 2023-04-01
Data refactoring

lymereJ:
This pull request sets up the new data management process for OpenStudio-Standards. The data management process now relies on a relational database which along some export functionalities enables developers and users of the repository to better track changes to the data, automatically validate data during updates, and maintain important relationship across different interdependent data tables (constructions/materials, space type assumptions, etc.). This provides a more sustainable data management approach and make it easier to expand upon (i.e., add data for additional building energy code; currently the database only covers ASHRAE 90.1-2004 through 2019).

This data refactoring effort also includes a refactoring of the space type data itself. Previously, space type data was defined as a function of building energy code version and building type which based on stakeholder surveys seemed confusing and complicated. The space type data has now been simplified and streamlined. The database offers users and developers a curated (concise yet exhaustive) list of space type to choose from. These space types and associated assumptions can be used to represent most space types in commercial building built in the U.S. 

A documentation of the new data management process is provided [here](https://github.com/NREL/openstudio-standards/blob/data_refactoring/data/openstudio_standards_data/README.md).

We expect this PR to stay open and serve as a reference for the upcoming workflow refactoring. We expect to have to provide modification to some of the proposed changes to accommodate changes to the workflow.

A few notes about the changes:
- A refactoring of the modeled fenestration assembly was included. Previously, some of the constructions were relying on the simplified glazing approach while others relied on a more detailed approach. All fenestration constructions have been switched to use the simplified glazing approach which is better suited to building energy code requirement which do not represent a specific product type
- Jenkins testing have been temporarily deactivated
- As per the documentation, test for the data management process are carried out using GitHub actions
- Since the number of changes provided in this PR is rather large, the data that can be exported from the database have not been included. They can be quickly and easily generated using the code provided in the documentation

PR #1550, opened 2023-07-26
reorganize the function

weilixu:


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -980,28 +980,39 @@ def model_apply_multizone_vav_outdoor_air_sizing(model)
...
+      #      air_loop_hvac_include_evaporative_cooler?(air_loop)) ||
+      #       (!air_loop_hvac_include_cooling_coil?(air_loop) &&
+      #         air_loop_hvac_include_economizer?(air_loop))
+      #      air_loop.additionalProperties.setFeature('non_mechanically_cooled', true)
+      #      air_loop.thermalZones.each do |thermal_zone|
+      #        thermal_zone.additionalProperties.setFeature('non_mechanically_cooled', true)
+      #      end
+      #    end
+      #  end
+      # end
```
lymereJ, line 1015: We probably don't want to keep the previous code as comments.

data/openstudio_standards_data/test/database_files_from_json/system_requirement_energy_recovery_90_1.json
```diff
@@ -0,0 +1,7346 @@
+[
```
lymereJ, line 1: Why is this included? This seems to be something from #1486.

lymereJ (approved):
I'll approved it but I think that the two small comments below should be addressed prior to merging.

PR #1552, opened 2023-08-01
add exterior lighting test function

weilixu:
reorganize the exterio lighting function to split the user data reading from the PRM data setting.

### DIFF
--- a/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
+++ b/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
@@ -709,69 +709,17 @@ class ASHRAE901PRM < Standard
   #
   # @param model [OpenStudio::model::Model] OpenStudio model object
   def model_apply_baseline_exterior_lighting(model)
-    user_ext_lights = @standards_data.key?('userdata_exterior_lights') ? @standards_data['userdata_exterior_lights'] : nil
-    return false if user_ext_lights.nil?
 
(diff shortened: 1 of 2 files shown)


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -709,69 +709,17 @@ def model_apply_prm_baseline_skylight_to_roof_ratio(model)
...
 
+    model.getExteriorLightss.each do |ext_lights_obj|
       # Update existing exterior lights object: control, schedule, power
       ext_lights_obj.setControlOption('AstronomicalClock')
       ext_lights_obj.setSchedule(model.alwaysOnDiscreteSchedule)
       ext_lights_obj.setMultiplier(1)
       ext_lights_def = ext_lights_obj.exteriorLightsDefinition
-      ext_lights_def.setDesignLevel(ext_ltg_pwr)
+      ext_ltg_pwr = get_additional_property_as_double(ext_lights_obj, 'design_level', 0.0)
+      if ext_ltg_pwr > 0.0
```
lymereJ, line 720: Is there a chance that `ext_ltg_pwr` be 0? If so, should this be greater or equal to?

lymereJ (approved):

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
weilixu: You are correct. it seems the ext_ltg_pwr is well protected from all the functions to be sure it is handled as 0.0 when failed.

In that case, we assume 0.0 and issue a warning. I will add these pieces in the code.

weilixu: Address comments

PR #1553, opened 2023-08-03
reorganize the elevator function and enhance the exceptional handling

weilixu:


lib/openstudio-standards/utilities/assertion.rb
```diff
@@ -62,6 +62,21 @@ def get_additional_property_as_boolean(component, key, default = false)
   return value
 end
 
+# PRM get an additional property from an OpenStudio object as a double,
+# if no such additional property, then return default value.
+# @param component [OpenStudio object] the component to get the additional property from
+# @param key [String] key string
+# @param default [Boolean] the default to return when there is no matching key
```
lymereJ, line 69: We should probably add the documentation for the return.

lymereJ (approved):

lib/openstudio-standards/utilities/assertion.rb
weilixu: Updated

PR #1561, opened 2023-08-09
remove VRP default in VAV constructor

mdahlhausen:
Removes model_system_outdoor_air_sizing_vrp_method(air_loop) and air_loop_hvac_apply_vav_damper_action(air_loop) methods from the default constructor for VAV systems. Instead, default to ZoneSum for the controller mechanical ventilation, and single maximum for VAV reheat terminals.  Still apply these methods before the sizing run in the create PRM baseline and create prototype methods.

Also, harmonize the outdoor air system setup between the VAV and PVAV archetypes.

Because of some slight name changes, there may be some adjustments to the regression models, though the performance results should not have changed.


mdahlhausen:
@lymereJ could you review this too? There should (hopefully) be no impact on the PRM methods.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -3249,7 +3249,7 @@ def model_add_crac(model,
         fan.setAvailabilitySchedule(hvac_op_sch)
       else
         OpenStudio.logFree(OpenStudio::Error, 'openstudio.Model.Model', "Fan type '#{fan_type}' not recognized, cannot add CRAC.")
-        return []
+        return false
```
lymereJ, line 3252: Looks like YARD comment for return should be updated.

lymereJ (approved):
If PRM tests pass, I'm fine with these changes. I just left a very minor comment.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
mdahlhausen: it's behind an error message, so the return type doesn't really matter. I'm aligning with the rest of the to return false after an error.

mdahlhausen:
Thanks for the review @lymereJ! Looks like there are a few prototype regression changes, but no PRM test failures.

PR #1566, opened 2023-08-11
Incorporating changes from nrcan branch

ckirney:


lib/openstudio-standards/standards/necb/ECMS/hvac_systems.rb
```diff
@@ -255,6 +255,7 @@ def get_max_vrf_pipe_lengths(model)
       end
     end
     max_net_vert_distance = max_vert_distance + min_vert_distance
+    max_net_vert_distance = [max_net_vert_distance, 0.000001].max
```
mdahlhausen, line 258: this line was deleted on master and should be deleted per a conversation with Kamel.  It was deleted on master; not sure how it got back in.

lib/openstudio-standards/standards/necb/NECB2011/necb_2011.rb
```diff
@@ -1,6 +1,9 @@
 # This class holds methods that apply NECB2011 rules.
 # @ref [References::NECB2011]
 class NECB2011 < Standard
+  require 'zip'
+  require 'open-uri'
```
mdahlhausen, line 5: do not add new require statements in the code

lib/openstudio-standards/standards/necb/ECMS/hvac_systems.rb
ckirney: Sorry, that what me.  I'll remove it.

ckirney: This should be fixed now.

lib/openstudio-standards/standards/necb/NECB2011/necb_2011.rb
mdahlhausen: Which gems require them? I did a `bundle install` and `gem list` and don't see them.

mdahlhausen: So I checked - open-uri was added to the ruby standard library in Ruby 3.0, and zip or rubyzip are not in the standard libraries. I think your options are to switch libraries, or add it to the gemspec. 

It looks like `uri` and `zlib` are default libraries in Ruby 2.7 that you could use instead.

mdahlhausen: Hmmm... after testing, it looks like 'open-uri' and 'zip' are built into the OS 3.6.1 docker container. So I think this is actually good as is.

mdahlhausen:
Also, it would help where applicable to only update regression tests for models that changed.  Some changes (version update, node naming as per the above comment) will change all the models. But the pump update didn't seem universal. That will help keep track of which changes actually changed a specific model, rather, than it just being part of a general update.

ckirney:
> Also, it would help where applicable to only update regression tests for models that changed. Some changes (version update, node naming as per the above comment) will change all the models. But the pump update didn't seem universal. That will help keep track of which changes actually changed a specific model, rather, than it just being part of a general update.

I understand the pain.  For small changes not affecting many tests we can try.  For changes that affect many tests we likely will continue to update all of them.

ckirney:
Thanks for the tips.  I think I addressed most of your concerns.  Let me know if I missed something or if there is anything else I should do.  Also, assuming you merge everything could you not delete the 'nrcan_master_merge' branch?  I'll integrate it into 'nrcan' branch (along with any other recent changes in master) when you are done.

mdahlhausen:
Thanks @ckirney! That addresses all my comments. I'll merge assuming the tests pass, and I'll leave the nrcan_master_merge open.

PR #1568, opened 2023-08-11
Feature/radiant controls 2 pipe plant

carlosduarteroa:
Adds support for two-pipe system based on outdoor air temperature or zone heating and cooling needs. #1540


mdahlhausen (changes requested):
adjust the test helper method and add tests for this in test_radiant_controls.rb

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -5769,6 +5787,257 @@ def model_cw_loop_cooling_tower_fan_type(model)
...
+  def model_two_pipe_water_plant(model, hot_water_loop, chilled_water_loop,
+                                 heating_plant_lockout: 65,
+                                 use_zone_demand: false,
+                                 thermal_zones: nil)
+
+
+      # create outdoor sensor node to be used in plant availability managers if needed
+      outdoor_airnode = model.outdoorAirNode
+      unless outdoor_airnode.initialized
+        require 'pry'; binding.pry
```
mdahlhausen, line 5830: no require statements

mdahlhausen:
@carlosduarteroa is this possible to accomplish with the SetpointManagerOutdoorAirReset availability manager? Maybe instead of modeling separate loops, we add one loop that is both the heating and cooling loop control by the OA reset manager. Then have that loops serve both heating/cooling side of the radiant objects.

mdahlhausen:
#1586 merged instead

PR #1570, opened 2023-08-14
Fix/vav vrp fix performance test changes

mdahlhausen:


mdahlhausen:
@lymereJ let me know if these changes look ok.
I also added the script to the repo, with some minor changes

lymereJ:
@mdahlhausen - Should we keep the script but revert the test results?

mdahlhausen:
Move script to #1574; don't commit regression/performance results changes

Issue #1539, opened 2023-07-21
Radiant Controls - Have the option to use zone occupancy object to determine the start and end of occupancy for the zone


mdahlhausen:

mdahlhausen:
Added with #1571

PR #1571, opened 2023-08-14
Feature/radiant controls zone occupancy

carlosduarteroa:
Adds support for using specific zone occupancy schedules for radiant control. #1539


mdahlhausen (changes requested):
Looks good; please adjust the test helper method and add tests for this in test_radiant_controls.rb

lib/openstudio-standards/prototypes/common/objects/Prototype.radiant_system_controls.rb
```diff
@@ -250,44 +263,105 @@ def model_add_radiant_proportional_controls(model, zone, radiant_loop,
     zone_rad_heat_operation_trend.setName("#{zone_name}_rad_heat_operation_trend")
     zone_rad_heat_operation_trend.setNumberOfTimestepsToBeLogged(zone_timestep * 48)
 
+    # use zone occupancy objects for radiant system control if selected
+    if use_zone_occupancy_for_control
```
mdahlhausen, line 267: I recommend implementing this with EnergyManagementSystemSensor on the Schedule value, like you have for the temperature setpoint schedules.

Use the schedule value to determine the control, rather than `IF ((CurrentTime >= #{zone_occ_hr_start_name}) && (CurrentTime <= #{zone_occ_hr_end_name})),`

That would mean changing it so if occupancy control isn't defined, and instead you get starting/stopping hours, making a ruleset schedule that has those hours and then following that schedule value for control.

To summarize:
1) create an on/off schedule from the zone occupancy schedule
2) if not zone control, create the on/off schedule from the provided start/stop times
3) track that zone schedule with a schedule_value EMS sensor
4) update the logic to check against the sensor value, rather than using  `IF ((CurrentTime >= #{zone_occ_hr_start_name}) && (CurrentTime <= #{zone_occ_hr_end_name})),`

mdahlhausen (changes requested):
I made a few changes (so pull again) and proposed a different method that is more closely tied to the zone occupancy schedule - and allows much more schedule variation.

lib/openstudio-standards/prototypes/common/objects/Prototype.radiant_system_controls.rb
mdahlhausen: - Use `thermal_zone_get_occupancy_schedule(zone, sch_name, occupied_percentage_threshold)` to make an on/off schedule
- name it `"#{zone.name} Radiant System Control Schedule"`
- add `occupied_percentage_threshold` argument to radiant controls method, and pass into `thermal_zone_get_occupancy_schedule`
- use the schedule value to control the system

mdahlhausen: Use `schedule_ruleset_set_hours_of_operation` to set schedule start/stop times

carlosduarteroa: I've made the changes you suggested with commit [54dd618](https://github.com/NREL/openstudio-standards/pull/1571/commits/54dd618273afc0f94de15172a7483f70444ecc73) and [b471c4a](https://github.com/NREL/openstudio-standards/pull/1571/commits/b471c4a2226d4db337200cb512ad9a444cf890be)

mdahlhausen (approved):
Nice work! I added a test for the whole building schedule option to the tests.

PR #1572, opened 2023-08-14
add test helper for radiant system controls

mdahlhausen:


mdahlhausen:
@carlosduarteroa Here is a default helper and some example tests for the radiant controls. You can pass in arguments to the test as if you were passing them to the `model_add_low_temp_radiant` method.

PR #1578, opened 2023-08-16
Fix for infiltration function.

weilixu:


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.rb
```diff
@@ -18,7 +18,7 @@ def load_standards_database(data_directories = [])
   #
   # @return [Double] the baseline infiltration rate, in cfm/ft^2
   # defaults to no infiltration.
-  def space_infiltration_rate_75_pa
+  def prm_space_infiltration_rate_75_pa
```
lymereJ, line 21: I'm not sure if this name change is useful since the function is already in a PRM specific class.

lib/openstudio-standards/utilities/assertion.rb
```diff
@@ -40,12 +40,24 @@ def prm_get_optional_handler(component, log_dir, data_key, *remaining_keys)
...
+            "Failed to retrieve data: #{data_key} from #{prm_get_component_name(component)}"
             )
   target_data_get = target_data.get
   return remaining_keys.length == 0 ? target_data_get : prm_get_optional_handler(target_data_get, log_dir, remaining_keys[0], *remaining_keys[1...])
 end
 
+# This is a PRM handler to get a name from an OpenStudio object instance
+# If the object instance does not have a name, then it will return the object name.
+#
+# @param component [OpenStudio] an OpenStudio object
```
lymereJ, line 52: We might want to return in the doc too.

lymereJ, line 52: I'm not sure if what's in the `[]` is the proper way of documenting an OpenStudio object.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -326,12 +327,13 @@ def model_current_building_envelope_infiltration_at_75pa(model, building_envelop
...
+  def model_adjusted_building_envelope_infiltration(building_envelope_area_m2)
     # Determine the total building baseline infiltration rate in cfm per ft2 of the building envelope at 75 Pa
-    basic_infil_rate_cfm_per_ft2 = space_infiltration_rate_75_pa
+    basic_infil_rate_cfm_per_ft2 = prm_space_infiltration_rate_75_pa
 
-    # Do nothing if no infiltration
-    return 0.0 if basic_infil_rate_cfm_per_ft2.zero?
+    # Do nothing if no infiltration - this is impossible as prm_space_infiltration_rate_75_pa always return 1.0
+    # remove
+    # return 0.0 if basic_infil_rate_cfm_per_ft2.zero?
```
lymereJ, line 336: We should probably remove these comments

lymereJ (approved):
This is good, I just left a few comments. I assume that the tests will pass so I'll approve but we should perhaps wait for the results before merging.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.rb
weilixu: The initial function name is overlapping with the function in [space_infiltration_rate_75_pa](https://github.com/NREL/openstudio-standards/blob/d0e3fabaeda812213835d134cd03feacec04ce4d/lib/openstudio-standards/standards/Standards.Space.rb#L1345) in the Standards.Space.rb.

The confusion becomes the `space_infiltration_rate_75_pa` in Standards.Space.rb has one argument but this function does not. Ruby also does not supporting function overloading so I worried that this might not right Ruby way.

Rubymine did called out the missing argument error though no running errors when complied to run.

Let me know if the original function name is intentional.

lib/openstudio-standards/utilities/assertion.rb
weilixu: Added, thanks!

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.rb
lymereJ: I think that we could add the argument to be consistent with the methods in Standards. The argument doesn't seem to actually be used, it just seemed to be added because the methods is in `***.Space.rb`.

lib/openstudio-standards/utilities/assertion.rb
weilixu: I am not sure too. I checked the API, it seems Component might be the correct one:

[Component](https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-3.6.1-doc/model/html/classopenstudio_1_1model_1_1_component.html) is a [Model](https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-3.6.1-doc/model/html/classopenstudio_1_1model_1_1_model.html) that contains exactly one [ComponentData](https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-3.6.1-doc/model/html/classopenstudio_1_1model_1_1_component_data.html) object.

All objects other than the [ComponentData](https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-3.6.1-doc/model/html/classopenstudio_1_1model_1_1_component_data.html) object and the version object are the [Component](https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-3.6.1-doc/model/html/classopenstudio_1_1model_1_1_component.html) contents. There is a primary [ModelObject](https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-3.6.1-doc/model/html/classopenstudio_1_1model_1_1_model_object.html) that provides the [Component](https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-3.6.1-doc/model/html/classopenstudio_1_1model_1_1_component.html)'s type. There are no public constructors, as Components are to be constructed through the [ModelObject](https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-3.6.1-doc/model/html/classopenstudio_1_1model_1_1_model_object.html) method [createComponent](https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-3.6.1-doc/model/html/classopenstudio_1_1model_1_1_model_object.html#a1da254fca2b648896a8c418985d86598). Similarly, Model-level object addition and removal, and SqlFile functionality, has been disabled.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/ashrae_90_1_prm_2019.rb
weilixu: That is OK with me too.
One thing I am not clear is though the function in `***.Space.rb` has a `Space` argument. But none of the implementations, including overriding functions use this argument.

My suggestion will be either remove this space argument in the parent class, or set the space to a default nil. I set the space to nil in the second implementation.

lymereJ: @mdahlhausen - Is there a reason for that (`space` argument for the `space_infiltration_rate_75_pa` method is provided but not used)?

mdahlhausen: I think it's just to align with the naming convention `object_some_method(object)`.

We don't need it; I'm fine to get rid of the space argument.  Or you could just pass in nil wherever you call it in the PRM workflow.

Also, in response to the above comments, ruby does overload functions - that how most of the standards code works - loading initial methods, then overloading them for each standard.

weilixu: @mdahlhausen Thanks, I put nil in case in the future Space is needed.

Most cases I see in the code base is overriding the function. Could you show me an example of such overloading in the code base? From what I read, Ruby does not support [function overloading](https://www.geeksforgeeks.org/method-overloading-in-ruby/).

mdahlhausen: Overloading - you are right in that ruby doesn't support argument overloading, but you can overload the method with the same arguments.

Glad the nill solution will work fine.

PR #1580, opened 2023-08-23
Feature/radiant controls supply water temperature based on outdoor air and load demand

carlosduarteroa:
These commits are to support #1542 (swt based on OA) and #1545 (swt based on zone heating and cooling needs). I still need to add unit tests.


mdahlhausen:
@carlosduarteroa does this PR require the occupancy and 2-pipe system PRs to function? If not, it'd be helpful to break it out separately.

mdahlhausen:
Replaced with #1595

PR #1581, opened 2023-08-25
Fix inheritance

lymereJ:


weilixu (approved):
Thank you for addressing the inheritance issue.

Issue #1441, opened 2023-03-14
Consolidate and remove duplicated 90.1 PRM methods


mdahlhausen:
Going through the code I noticed that several of the ASHRAE 90.1 PRM methods are functionally identical duplicates of methods already in standards.

For example, the hot water boiler methods [`boiler_hot_water_find_search_criteria`](https://github.com/NREL/openstudio-standards/blob/61214a372d74b219bbacef53f665e3ffd4903afa/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.BoilerHotWater.rb) and `boiler_hot_water_standard_minimum_thermal_efficiency` are functionally identical to the [generic Standards versions](https://github.com/NREL/openstudio-standards/blob/73d09fff27ac00da15d521bc817092159b6f3c0d/openstudio-standards/lib/openstudio-standards/standards/Standards.BoilerHotWater.rb).

There are other instances where including an extra optional argument would prevent having to overload the whole method with a specific 90.1 PRM version.

In the interest of keeping the codebase smaller and maintainable, this issue is to consolidate and remove such methods.

mdahlhausen:
Copying @lymereJ's comments on duplicated PRM methods:

- [ ] I did not see the “model_custom_hvac_tweaks” methods and I guess more broadly all the methods under prototypes\common\building\, we are going to remove them all, right?
- [ ] deep_copy_schedule  from ashrae_90_1_prm.SpaceType.rb should probably moved to the utility class that you’re working on
- [x] I think that the PRM specific “chw_sizing_control” could be revised so we only have one method, it looks like the differences between this one and the one that’s in Standards are default values and the fact that we’re using an OA reset SPM which I see is in a todo note in the Standards version.
- [ ] “planar_surface_apply_standard_construction” could probably be one method with some different references for the PRM, that would save a bit of duplicated code.
- [x] “model_prm_baseline_system_group_minimum_area” is an interesting one. The PRM specific one has a custom flag for Xcel Energy, but not the standard one. I wonder if these have been inverted? Either way, one of them could be removed since the exception area appears to be the same.
- [x] Same thing for “model_prm_baseline_system_number” and “model_prm_baseline_system_change_fuel_type”, although the content appears to be different. Maybe they need to be kept, but the Xcel energy custom flag should probably be moved to the standard one.
- [ ] We can probably have one “model_apply_constructions” since the only difference is the default values passed to “model_apply_standard_constructions”
[...]

PR #1583, opened 2023-08-28
Remove duplicate prm methods

mdahlhausen:
Partially addresses #1441.

Notes: 'fan_type' is defined as 'Propeller or Axial' in Standards.CoolingTower.rb, and 'Axial' in the PRM version. The PRM data should switch to 'Propeller or Axial' to align with 90.1. That's a separate commit to author and add to this PR.


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.CoolingTower.rb
```diff
@@ -1,110 +0,0 @@
...
-    # in the hvac standards data set.
-    search_criteria = {}
-    search_criteria['template'] = template
-
-    # By definition cooling towers in E+ are open.
-    # Closed cooling towers are the fluidcooler objects.
-    search_criteria['equipment_type'] = 'Open Cooling Tower'
-
-    # Define the criteria to find the fan type
-    search_criteria['fan_type'] = 'Axial'
```
weilixu, line 48: I generally agree this function is largely duplicate with standard.
To remove this function, we should also update the [data file](https://github.com/NREL/openstudio-standards/blob/AppendixG_Dev/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm_2019/data/ashrae_90_1_prm_2019.heat_rejection.json#L6) to correctly extract the performance.

Once the data file is corrected, the hvac_efficiency test can pass the CI

PR #1588, opened 2023-08-31
Appx g/bugfix/duplicate ltg schedules

weilixu:
Fix issues in lighting schedule duplication and missing lighting power density when standard space type is not one of the PRM standard space type.


weilixu:
Apologize for mixing this up with YARD comment PR. Once the YARD comment PR is merged, this should only showing the changes for the lighting power density and schedule.

PR #1611, opened 2023-09-19
Appx g/feature/export user data partial

weilixu:
Pull request overview
---------------------
Add a function to generate user data template from a user model.

 - Fixes #ISSUENUMBERHERE (IF THIS IS A DEFECT)

### Pull Request Author

 - [x] Method changes or additions
 - [ ] Data changes or additions
 - [ ] Added tests for added methods
 - [ ] If methods have been deprecated, update rest of code to use the new methods
 - [x] Documented new methods using [yard syntax](https://rubydoc.info/gems/yard/file/docs/GettingStarted.md)
 - [ ] Resolved yard documentation errors for new code (ran `bundle exec rake doc`)
 - [ ] Resolved rubocop syntax errors for new code (ran `bundle exec rake rubocop`)
 - [ ] All new and existing tests passes
 - [ ] If the code adds new `require` statements, ensure these are in core ruby or add to the gemspec

### Review Checklist

This will not be exhaustively relevant to every PR.
 - [ ] Perform a code review on GitHub
 - [ ] All related changes have been implemented: method additions, changes, tests
 - [ ] Check rubocop errors
 - [ ] Check yard doc errors
 - [ ] If fixing a defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If a new feature, test the new feature and try creative ways to break it
 - [ ] CI status: all green or justified

PR #1608, opened 2023-09-15
Appx g/enhancement/restructure user data

weilixu:
Pull request overview
---------------------

Restructure the user data.
Explicitly write out the data requirements for some of the fields in the implemented user data.
More to add in the future.

Separately the responsibility in three aspects:
1. User data loading (validation functions)
2. User data handling (handle data for model processing)
3. Get additionalFeatures (processed data)

### Pull Request Author

 - [x] Method changes or additions
 - [x] Data changes or additions
 - [ ] Added tests for added methods
 - [ ] If methods have been deprecated, update rest of code to use the new methods
 - [x] Documented new methods using [yard syntax](https://rubydoc.info/gems/yard/file/docs/GettingStarted.md)
 - [ ] Resolved yard documentation errors for new code (ran `bundle exec rake doc`)
 - [ ] Resolved rubocop syntax errors for new code (ran `bundle exec rake rubocop`)
 - [ ] All new and existing tests passes
 - [ ] If the code adds new `require` statements, ensure these are in core ruby or add to the gemspec

### Review Checklist

This will not be exhaustively relevant to every PR.
 - [ ] Perform a code review on GitHub
 - [ ] All related changes have been implemented: method additions, changes, tests
 - [ ] Check rubocop errors
 - [ ] Check yard doc errors
 - [ ] If fixing a defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If a new feature, test the new feature and try creative ways to break it
 - [ ] CI status: all green or justified

Issue #1527, opened 2023-06-07
90.1 PRM: Redundant SpaceType Lights schedule created for each Space in SpaceType


eringold:
@dmaddoxwhite @lymereJ @weilixu 

I'm getting multiple redundant lighting schedules created in the stable PRM baseline model, it looks like one per instance of Space in each SpaceType:
![image](https://github.com/NREL/openstudio-standards/assets/10799524/9367b40c-1062-4fb7-a29f-1107d1a12114)

I think this is due to the logic in ashrae_90_1_prm.SpaceType `space_type_light_sch_change`. For each space, the SpaceType lights schedule is retrieved, then the name changed, then it checks whether the _new_ name is in `schedule_hash`, and if not the schedule is copied and assigned again to the Lights, and the new name added to `schedule_hash`. Because the name is changed before it's checked, the `else` branch will run and a new copy will always happen for each space:

https://github.com/NREL/openstudio-standards/blob/12bbfabf3962af05b8c267c1da54b8e3a89217a0/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb#L440-L468

Basically I think https://github.com/NREL/openstudio-standards/blob/12bbfabf3962af05b8c267c1da54b8e3a89217a0/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb#L453

should be 
```
if schedule_hash.key?(ltg_schedule_name)
```

weilixu:
@eringold thanks for reporting. This issue will be addressed in the next release.

PR #1407, opened 2023-01-13
Appendix g dev debug user directory error

weilixu:
The command line interface is failing when calling create PRM method.
Failing message indicates the directory for user data files are arrays instead of strings.

Fix the issue in the convert_user_data_to_json function, and verified the workflow in the command line.

PR #1612, opened 2023-09-22
Appendix g dev

weilixu:
Pull request overview
---------------------
Merge Appendix G PRM development.

Features: 
1. #1514 Add baseline outdoor air workflow
2. #1556 Add the capability to generate a proposed building model from a user model - this includes adjust infiltration values and lighting power density for dwelling units and model/hotel, equipment schedule for computer space.
3. #1611 Add export user data capability that can export a user data template based on user model.

Enhancements:
1. #1507 Handle missing standard space type in user model
4. #1514 Add OSSTD-PRM exception handler module to handle 1. get OpenStudio option objects, 2. handle reading the user data from csv file, 3. handle getting missing additional properties.
5. #1533 Update the function `run_four_orientation` using methods in exception handler module.
6. #1549 Update the function `update_ground_temperature_profile` using methods in exception handler module.
7. #1550 Update the function `model_identify_non_mech_cooled_systems` using methods in exception handler module.
8. #1552 Update the function `apply_baseline_exterior_lighting` using methods in exception handler module, move user data to user data process function
9. #1553 Update the function `model_add_prm_elevators` using methods in exception handler module.
10. #1608 Restructure user data module to create a enum library, a validation workflow and a data processing workflow.

Bug fixs:
1. #1405 Unit test: handling unsuccessful proposed model generation
12. #1407 Fix reading user data directory error
13. #1428 Resolve infinite loop in unmet load hour check and debug SQL session block in Windows OS.
14. #1588 Address the issue described in #1527 - spaceType lights schedule duplicated.
15. #1589 Address fenestration missing constructions when apply PRM method
16. #1610 Fix generating water heating coil in PSZ-VAV system causing EnergyPlus crash for duplicated components error.

### Pull Request Author

 - [x] If the code adds new `require` statements, ensure these are in core ruby or add to the gemspec

### Review Checklist

 - [x] Check rubocop errors
 - [x] Check yard doc errors


weilixu:
@mdahlhausen I am getting multiple invalid tag format, for example: 
**[warn]: Invalid tag format for @return in file `lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb` near line 380**

Here is the line 379 @return comment:
`  # @return space_array [Array] List of Spaces [OpenStudio::Model::Space]`

I am not sure what could be wrong in this case. Do you know how to fix this error message?

lib/openstudio-standards/standards/Standards.Space.rb
```diff
@@ -2983,4 +2983,12 @@ def space_daylighting_control_type(space)
   def space_daylighting_minimum_input_power_fraction(space)
     return 0.3
   end
+
+  # Create and assign PRM computer room electric equipment schedule
+  #
+  # @param space [OpenStudio::Model::Space] OpenStudio Space object
+  # @return [Boolean] returns true if successful, false if not
+  def space_add_prm_computer_roomm_equipment_schedule(space)
```
mdahlhausen, line 2991: typo in method name (roomm)

lib/openstudio-standards/standards/Standards.PlantLoop.rb
```diff
@@ -1495,4 +1495,43 @@ def plant_loop_capacity_w_by_maxflow_and_delta_t_forwater(plant_loop)
...
+  # @param plant_loop [OpenStudio::Model::PlantLoop] plant loop
+  # @return [Bool] returns true if successful
+  def plant_loop_adiabatic_pipes_only(plant_loop)
+    supply_side_components = plant_loop.supplyComponents
+    demand_side_components = plant_loop.demandComponents
+    plant_loop_components = supply_side_components += demand_side_components
+    plant_loop_components.each do |component|
+      # Get the object type
+      obj_type = component.iddObjectType.valueName.to_s
+      next if !['OS_Pipe_Indoor', 'OS_Pipe_Outdoor'].include?(obj_type)
```
mdahlhausen, line 1511: check rubocop on this - think it should be next unless, not next if !

lib/openstudio-standards/utilities/assertion.rb
```diff
@@ -0,0 +1,135 @@
...
+# This is a PRM handler function handles the .get from an optional object
+# The handler will try to access the OpenStudio Object data key
+# And do it recursively until all the keys have been checked and final object get or raise exception
+# for non_initialized objects.
+# @param component [OpenStudio::Model::Component] an OpenStudio object
+# @param log_dir [string] directory to save the log
+# @param data_key [string] The data key to retrieve the data from the OpenStudio object
+# @param remaining_keys [str] Any additional keys in the path
+# @return the OpenStudio Object or exception raise
+def prm_get_optional_handler(component, log_dir, data_key, *remaining_keys)
```
mdahlhausen, line 39: Why is this function necessary?

lib/openstudio-standards/standards/ashrae_90_1_prm/userdata_csv/userdata_enums.rb
mdahlhausen, line 1: why the use of global variables instead of instance variables for class UserData?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Surface.rb
```diff
@@ -8,7 +8,7 @@ class ASHRAE901PRM < Standard
   def surface_adjust_fenestration_in_a_surface(surface, reduction, model)
     # do nothing for cases when reduction == 1.0
     if reduction < 1.0
-      surface.subSurfaces.sort.each do |ss|
+      surface.subSurfaces.each do |ss|
```
mdahlhausen, line 11: There are some pieces of code which give different results depending on whether .sort is called or not. If the object is being removed, you probably don't need it, but it may be worth including the sort to be safe.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb
```diff
@@ -614,6 +650,7 @@ def has_multi_lpd_values_user_data(user_data, space_type)
   # The sum of each space fraction in the user_data is assumed to be 1.0
   # @param user_data [Hash] user data from the user csv
   # @param space [OpenStudio::Model::Space]
+  # @return space_lighting_per_area [Float]
```
mdahlhausen, line 653: syntax is @return [Float] value units. run `bundle exec rake doc` and see errors

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb
```diff
@@ -578,15 +608,21 @@ def has_user_lpd_values(user_space_data)
...
+
     return space_type.spaces.size > 1 && lighting_per_length > 0
   end
 
   # Function checks whether there are multi lpd values in the space type from user's data
   # The sum of each space fraction in the user_data is assumed to be 1.0
   # multi-lpd value means lighting per area > 0 and lighting_per_length > 0
+  # @param user_data [Hash] user data from the user csv
+  # @param space_type [OpenStudio::Model::SpaceType]
+  # @return has_multi_lpd_values [Boolean]
```
mdahlhausen, line 625: @return [Boolean] value units

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb
```diff
@@ -578,15 +608,21 @@ def has_user_lpd_values(user_space_data)
 
   # Function checks whether there are multi lpd values in the space type
   # multi-lpd value means there are multiple spaces and the lighting_per_length > 0
+  # @param space_type [OpenStudio::Model::SpaceType]
+  # @return has_multi-lpd_values_space_type [Boolean]
```
mdahlhausen, line 612: @return [Boolean] value units

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb
```diff
@@ -567,6 +595,8 @@ def calculate_lpd_by_space(space_type, space)
   end
 
   # Function checks whether the user data contains lighting data
+  # @param user_space_data [Hash] space data extracted from user csv.
+  # @return has_user_lpd_values [Boolean] True if there are user lpd values, False otherwise.
```
mdahlhausen, line 599: @return [Boolean] value units

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb
```diff
@@ -532,11 +558,13 @@ def deep_copy_schedule(new_schedule_name, schedule, adjustment_factor, model)
   # calculate the lighting power density per area based on space type
   # The function will calculate the LPD based on the space type (STRING)
   # It considers lighting per area, lighting per length as well as occupancy factors in the database.
-  # @param space_type [String]
+  # @param space_type [OpenStudio::Model::SpaceType]
   # @param space [OpenStudio::Model::Space]
+  # @return space_lighting_per_area [Float] lighting power density in the space
```
mdahlhausen, line 563: @return [Float] value units

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb
```diff
@@ -364,9 +373,10 @@ def set_lpd_on_space_type(space_type, user_spaces, user_spacetypes)
   # This function is used when there are user space data available or
   # the spaces under space type has lighting per length value which may cause multiple
   # lighting power densities under one space_type.
-  # @param user_spaces hash data contained in the user space
-  # @param user_spacetypes hash data contained in the user spacetypes
-  # @param space_type OpenStudio::Model::SpaceType object
+  # @param user_spaces [Hash] hash data contained in the user space
+  # @param user_spacetypes [Hash] hash data contained in the user spacetypes
+  # @param space_type [OpenStudio::Model::SpaceType] object
+  # @return space_array [Array] List of Spaces [OpenStudio::Model::Space]
```
mdahlhausen, line 379: @return [Array<OpenStudio::Model::Space>] description

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb
```diff
@@ -158,6 +158,7 @@ def calculate_electric_value_by_userdata(user_equip_data, power_equipment, power
     # Check if the plug load represents a motor (check if motorhorsepower exist), if so, record the motor HP and efficiency.
     if !user_equip_data['motor_horsepower'].nil?
       # Pre-processing will ensure these three user data are added correctly (float, float, boolean)
+      # TODO move this part to user data processing.
```
mdahlhausen, line 161: todo syntax is @todo

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb
```diff
@@ -441,22 +451,38 @@ def space_type_light_sch_change(model)
     # set schedule for lighting
     schedule_hash = {}
     model.getSpaces.each do |space|
-      space_type = space.spaceType.get
-      if space_type.hasAdditionalProperties && space_type.additionalProperties.hasFeature('regulated_lights_name')
+      space_type = prm_get_optional_handler(space, @sizing_run_dir, 'spaceType')
+      if has_additional_feature(space_type, 'regulated_lights_name')
```
mdahlhausen, line 455: method abstraction unnecessary

lib/openstudio-standards/standards/Standards.PlantLoop.rb
```diff
@@ -1495,4 +1495,43 @@ def plant_loop_capacity_w_by_maxflow_and_delta_t_forwater(plant_loop)
...
+      pipe = nil
+      case obj_type
+      when 'OS_Pipe_Indoor'
+        pipe = component.to_PipeIndoor.get
+      when 'OS_Pipe_Outdoor'
+        pipe = component.to_PipeOutdoor.get
+      end
+
+      # Get pipe node
+      node = prm_get_optional_handler(pipe, @sizing_run_dir, 'to_StraightComponent', 'outletModelObject', 'to_Node')
```
mdahlhausen, line 1523: why not use the outletModelObject field from the StraightComponent parent class instead?

lib/openstudio-standards/utilities/assertion.rb
mdahlhausen, line 1: not sure the additional property methods are necessary. if you want to include them, they should be generic, and not specific to the prm logging. (change error message)

test/90_1_prm/test_appendix_g_prm.rb
mdahlhausen, line 1: I should have mentioned this last time:
Split out this file into a helper methods file, and a test file. The test_appendix_g_prm should only include the test_ methods. (It sometimes makes sense to break this practice, but I don't it makes sense here).

lib/openstudio-standards/utilities/object_info.rb
```diff
@@ -1,20 +1,41 @@
...
+    end
+    return output_hash
+  end
+
+  # This method is used to automatically cast a model
+  # object based on its IDD object type
+  #
+  # @param [OpenStudio::Model::ModelObject]
+  # @return Casted OpenStudio object or nil if the cast was not possible
+  def model_cast_model_object(model_object)
```
mdahlhausen, line 29: I think this method is unnecessary and could result in weird behavior, particularly for HVAC where it is called. I suggest refactoring to remove it in the two places it is used. Happy to think about how to refactor those methods.

lib/openstudio-standards/utilities/object_info.rb
```diff
@@ -1,20 +1,41 @@
...
+Standard.class_eval do
+  # A set of method to manipulate objects
+
+  # Get hash of object field names and values
+  # @author: Xuechen (Jerry) Lei, PNNL
+  # @param  obj [object]
+  #
+  # @return [Hash<String>] FieldName:Value
+  #
+  def model_get_object_hash(obj)
```
mdahlhausen, line 10: This method is never used? delete?

lib/openstudio-standards/utilities/object_info.rb
mdahlhausen, line 1: Can probably delete this file as methods are unused or could be refactored and embedded inline.

lib/openstudio-standards/utilities/assertion.rb
```diff
@@ -0,0 +1,135 @@
...
+end
+
+# PRM assertion method
+# Raise assertion if the test (bool) is failed
+#
+# Before raise the exception, the method will generate the prm.log for debugging
+# @param bool [Boolean, Object] an object for truthy evaluation
+# @param log_dir [String] log file directory
+# @param err_msg [String] message raise the exception
+def prm_raise(bool, log_dir, err_msg)
```
mdahlhausen, line 12: why a PRM specific error message?

lib/openstudio-standards/standards/Standards.Space.rb
weilixu: Resolved.

lib/openstudio-standards/utilities/assertion.rb
weilixu: We have many situations to get optional object from an OpenStudio object.
```Ruby
spacetype = space.spaceType.get.standardsSpaceType.get
```
In above code, if a space has no spaceType or standardsSpaceType, than ruby will raise exception with message `nil has no get function`. This happens quite often with models generated from BEM software. The error message is not informative.

The `prm_get_optional_handler` is designed to handle such scenario to give more informative error messages to user.

lib/openstudio-standards/standards/ashrae_90_1_prm/userdata_csv/userdata_enums.rb
weilixu: I thought about using global variables, but it loses some functionalities specific to a group of data.

For example, there are many flags in the user data that require user to type `true` or `false` in the CSV file.
To check whether user provided correct string to these flags, we need to have a way to verify those strings are not `yes` or `no` and they are case insensitive.

The advantage of class implementation is that the UserData class can have a generic `matched_any?` method which compares a string with a list of expected enums.
Then in the UserDataBoolean, subclass of UserData, we can specify the expected enums (`true` and `false`)

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Surface.rb
weilixu: Got it. Thanks for pointing it out. I will check and add them back.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.SpaceType.rb
weilixu: Agreed. Remove the `has_additional_feature`.

lib/openstudio-standards/utilities/assertion.rb
weilixu: I am not sure how useful these can be for other routines.
For PRM, it is a little different.

The way how we process the user data is:
1. Read in CSV, convert to Json, validate through validation module
2. Process validated user data, add information to additionalProperties if any.
3. In PRM method body, read these additionalProperties for prm specific process.

Often time, we see if user data is missing a data, the additionalProperties for that particular data is not added to the OpenStudio object, thus causing failure with Ruby default error message.


To make it generic, I will change the `prm.log` to `openstudio.model.addtionalproperties.`

lib/openstudio-standards/utilities/object_info.rb
weilixu: Removed

lib/openstudio-standards/standards/Standards.PlantLoop.rb
weilixu: updated.

lib/openstudio-standards/utilities/object_info.rb
lymereJ: This method is actually quite convenient. For example, with a few lines of code, it allow us to check if an HVAC related component in a model is autosized without actually knowing what its full object type name is in advance. Without it we would have to list in the code all the object that we want to check. The proposed approach also supports future objects to be included in OpenStudio/EnergyPlus (i.e., new coil objects). It has existed in the codebase for a while now, we just generalized it here to use it in multiple places.

What are the weird behavior that you'd expect to see? If you don't want to expose it we could make it private.

lib/openstudio-standards/standards/Standards.PlantLoop.rb
lymereJ: That lines is equivalent to `pipe.to_StraightComponent.outletModelObject.to_Node`, so we do use the `outletModelObject` from the `StraightComponent` parent class. Could you please clarify what you mean?

test/90_1_prm/test_appendix_g_prm.rb
weilixu: Make senses. Given the time constraint, could we move this to next release? I will open an issue in our backlog to address it.

lib/openstudio-standards/standards/Standards.PlantLoop.rb
mdahlhausen: just replace it with node = pipe.to_StraightComponent.outletModelObject.to_Node

also not sure node_name is used in the following lines?

lymereJ: Doesn't `to_Node` return an optional?

lib/openstudio-standards/utilities/assertion.rb
mdahlhausen: The better practice is to call .is_initialized each time, and then have comments at that point in code if the optional value is not initialized.

Search .is_initialized in the codebase and you'll see many examples of if - else statements to catch this sort of thing.

lib/openstudio-standards/utilities/assertion.rb
mdahlhausen: Thanks for explaining. I'm fine to keep the prm_raise method.

lib/openstudio-standards/utilities/assertion.rb
mdahlhausen: ok that's fine.

The error message should help point to the section of code with the error, so 'openstudio.standards.utilities'

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -1,4 +1,5 @@
 require 'csv'
+require 'date'
```
mdahlhausen, line 2: is this in the core or std library?

lib/openstudio-standards/utilities/assertion.rb
weilixu: Right, but there are multiple chain options in the codebase so this method comes in as convenience to handle those chain optional get.
I can add .is_initialized in the `prm_get_optional_handler` method to reflect the best practice?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -2024,32 +2110,35 @@ def model_update_ground_temperature_profile(model, climate_zone)
...
+      ground_temp = OpenStudio::Model::SiteGroundTemperatureFCfactorMethod.new(model)
+      ground_temperatures = model_get_monthly_ground_temps_from_stat_file(stat_file_path)
+      unless ground_temperatures.empty?
+        # set the site ground temperature building surface
+        ground_temp.setAllMonthlyTemperatures(ground_temperatures)
+      end
     end
 
-    return true
+    return surfaces_with_fc_factor_boundary
```
mdahlhausen, line 2141: update return type in method documetnation

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
```diff
@@ -2182,6 +2272,7 @@ def model_get_bat_wwr_target(bat, wwr_list)
   # @param total_fene_m2 [Float] total fenestration area
   # @return [Float] reduction factor
   def model_get_wwr_reduction_ratio(multiplier,
```
mdahlhausen, line 2274: method should be surface_get_wwr_reduction_ratio.
And then just pass in the surface. You pass in name, and results from methods with surface as an argument.

lib/openstudio-standards/utilities/assertion.rb
mdahlhausen: I see 20 uses of this method in the code base, and 1 instance of chaining, which I elsewhere commented wasn't necessary.
I think this method doesn't accomplish what you want - giving more clear error messages to the user.

lib/openstudio-standards/utilities/assertion.rb
weilixu: Sure will update it.

lib/openstudio-standards/standards/Standards.Model.rb
weilixu: I think this comes from Ruby standard library - see [Ruby 2.7.2](https://ruby-doc.org/stdlib-2.7.2/libdoc/date/rdoc/Date.html)

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
weilixu: The `surfaces_with_fc_factor_boundary` is a Boolean in the code.
I updated the comment to:
`# @return [Boolean] surfaces_with_fc_factor_boundary,  returns true if successful, false if not`

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
weilixu: Do you suggest to just pass in a surface and then get `surface_name`, `surface_wwr` and `surface_dr` shall be a local variable inside the `model_get_wwr_reduction_ratio`?

test/90_1_prm/test_appendix_g_prm.rb
mdahlhausen: it's simple:
1) move non-test methods to another file
2) add a require statement at the top of the test file

lib/openstudio-standards/utilities/assertion.rb
weilixu: In the PRM routine, there are growing number of chaining for processing user data including some of them pointing to `name`.
Examples include:
[Get building user data name](https://github.com/NREL/openstudio-standards/blob/AppendixG_Dev/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb#L1748) 
[get space types:](https://github.com/NREL/openstudio-standards/blob/AppendixG_Dev/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb#L2403)

Most of them are specific to PRM user data processing so I would like, if possible, to keep it for now.

I can update the error message to make the issue clearer to the user?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.Model.rb
weilixu: This function has updated to `surface_get_wwr_reduction_ratio` with `surface` as a new argument to replace `surface_name`, `surface_wwr` and `surface_dr`

test/90_1_prm/test_appendix_g_prm.rb
weilixu: Done, removed.
It could be further improved, I will put this work to our blacklog. Thanks @mdahlhausen

mdahlhausen (approved):

Issue #1510, opened 2023-05-10
move or reduce weather file size


mdahlhausen:
Perhaps there is a separate weather gem, or we reduced the weather files down to ~30 key ones.

NRCAN to come up with proposal.

mdahlhausen:
Significantly addressed in #1668.

Next step is to remove the excessive California weather files.

mdahlhausen:
California weather files removed in #1650

PR #1650, opened 2024-01-02
Refactor weather methods

eringold:
Pull request overview
---------------------

Adds weather-related information and modify methods, as well as classes for parsing EPW and STAT files. Includes all functionality of ChangeBuildingLocation measure to standards. 

### Pull Request Author

 - [x] Data changes or additions
 - [x] If methods have been deprecated, update rest of code to use the new methods

### Review Checklist

 - [x] Check rubocop errors
 - [x] Check yard doc errors
 - [x] If a new feature, test the new feature and try creative ways to break it


mdahlhausen (approved):
@eringold changes look good. I moved over more methods and deleted the old ones. I'm going to deal with Weather.Model.rb and the WeatherFile class in this PR before I merge it.

mdahlhausen:
commit cf84aa8 fixes #1510

mdahlhausen:
@srgilani Could you review the changes from these two commits?:
- [calculate design day average global irradiance](https://github.com/NREL/openstudio-standards/pull/1650/commits/12935ac4e42d4e5e82af2b1d6837a435ce023a63)
- [implementation of the dehumidification "degree days" method](https://github.com/NREL/openstudio-standards/pull/1650/commits/dbb4d36521ecfe9919097b182820280f349de7e8)

I can't quite recreate the values in Graham Wright's paper for average global irradiance. The values I do calculate are reasonable and accurate, but are bit higher than his range. The dehumidification degree days are roughly the same - perhaps a bit higher. I noticed your method would take the average for an entire day, and set ddds to zero if the average did not meet the 0.010 threshold. My method sums all hourly values over 0.010, then divides by 24 hours - similar to how one would calculate heating or cooling degree days.

lib/openstudio-standards/standards/necb/common/btap_data.rb
```diff
@@ -2089,11 +2089,25 @@ def phius_performance_indicators(model)
...
     ### IGHL (Irradiance, Global, at the heating design condition) (Btu/h.ft2) in PHIUS, 2021
-    solar_irradiance_on_heating_design_day_w_per_m_sq = BTAP::Environment::WeatherFile.new(weather_file).get_ghi_on_heating_design_day
+    average_daily_global_irradiance_w_per_m2_array = []
+    model.getDesignDays.each do |design_day|
+      next unless design_day.dayType == 'WinterDesignDay'
+
+      average_daily_global_irradiance_w_per_m2 = OpenstudioStandards::Weather.design_day_average_global_irradiance(design_day)
+      average_daily_global_irradiance_w_per_m2_array << average_daily_global_irradiance_w_per_m2
+    end
+    solar_irradiance_on_heating_design_day_w_per_m_sq = average_daily_global_irradiance_w_per_m2_array.min
```
srgilani, line 2099: I did not follow why min (or max) of "average_daily_global_irradiance_w_per_m2_array" is calculated here... wouldn't that be only one day which is WinterDesignDay (or SummerDesignDay)?

lib/openstudio-standards/weather/information.rb
```diff
@@ -362,6 +362,80 @@ def self.get_climate_zone_from_degree_days(hdd18, cdd10)
...
+          beam_normal_irradiance = 0.0
+          diffuse_horizontal_irradiance = 0.0
+        end
+        global_irradiance = beam_normal_irradiance + diffuse_horizontal_irradiance
+        global_irradiance_array << global_irradiance
+
+        # puts "For local_standard_time_hour #{local_standard_time_hour}, apparent_solar_time #{apparent_solar_time}, hour_angle_degrees #{hour_angle_degrees}, solar_altitude_degrees #{solar_altitude_degrees}, air_mass #{air_mass}, beam_normal_irradiance #{beam_normal_irradiance}, diffuse_horizontal_irradiance #{diffuse_horizontal_irradiance}"
+      end
+
+      average_daily_global_irradiance = global_irradiance_array.sum / 24.0
```
srgilani, line 429: I noticed you are averaging over 24 hours of the day. I am still not sure what PHIUS meant by that…

lib/openstudio-standards/weather/information.rb
```diff
@@ -436,6 +436,60 @@ def self.design_day_average_global_irradiance(design_day)
...
+
+      # calculate total pressure (Pa)
+      tp_values = pp_values.zip(atm_p_values).map{ |pp, atm| pp + atm }
+
+      # calculate humidity ratio
+      hr_values = pp_values.zip(tp_values).map{ |pp, tp| (0.621945 * pp) / (tp - pp)}
+
+      # calculate dehumidification degree days based on humidity ratio values above the base
+      hr_values_above_base = hr_values.map{ |hr| hr > base_humidity_ratio ? hr : 0.0}
+      dehumidification_degree_days = hr_values_above_base.sum / 24.0
```
srgilani, line 484: I tried to find any sources for it; the only thing I found is that Graham Wright says in his paper that “DDD was calculated by thresholding the monthly humidity data at a humidity ratio of 0.010”. To me it sounded like they considered monthly average, so I used it as daily humidity ratio… 
I tested for Montreal, the two methods give 0.76 using your method and 0.15 using my method… I’m not sure which one would make more sense for what PHIUS is asking for… we can use the same method as heating/cooling degree days as you suggested

srgilani (commented):
@mdahlhausen I reviewed your methods and went through ASHRAE Chapter 14 that you used for calculating global irradiance; agreed with your methods/changes. I had only a comment/question about your changes in btap_data.rb

ckirney:
I will make a branch and test with our costing stuff to make sure there are no issues.

lib/openstudio-standards/standards/necb/common/btap_data.rb
mdahlhausen: So what this is saying is for the heating condition - look at all the heating design days, and log the average daily global irradiance.
Then, take the minimum of these values, meaning use the one which has the lowest average daily global irradiance. This is the most extreme condition - heating with the least amount of solar gain.

The inverse is for cooling - we want the most extreme condition, which is cooling with the most amount of solar gain.

lib/openstudio-standards/weather/information.rb
mdahlhausen: Yes, I went with the daily average. It's both a bit more straightforward to calculate, and gives numbers closer to the range in Graham Wright's paper. Only during sun hours is both more complicated, and means that the average value across design days isn't comparable, because the sun hours aren't the same.

lib/openstudio-standards/weather/information.rb
mdahlhausen: If Graham is using monthly humidity averages, then both of our methods are wrong. It also wouldn't be "degree-days" in the normal meaning of the term. It's often the case that the sum of monthly heating degree days at 18C are non-zero when the monthly average temperature is above 18C.
I think the best way to resolve this is to write up and implement a proposal to PHIUS to implement a better and clearer regression. But I don't have time for that. Until then I suggest we settle with the methods we have that at least are clear and somewhat defensible.

ckirney (approved):
I noticed some small changes in results but it looks good to me.

mdahlhausen:
@ckirney thanks for the review! The changes are from adding the annual wetbulb condition. It seems that was intended to be added to the necb models, but the specific regex used wasn't adding it.

phylroy (approved):

PR #1656, opened 2024-01-10
Refactor/schedules module

mdahlhausen:
Pull request overview
---------------------

This PR moves several schedules methods from:
```
openstudio-standards-0.4.0/lib/openstudio-standards/standards/Standards.Model.rb
openstudio-standards-0.4.0/lib/openstudio-standards/standards/Standards.ScheduleCompact.rb
openstudio-standards-0.4.0/lib/openstudio-standards/standards/Standards.ScheduleConstant.rb
openstudio-standards-0.4.0/lib/openstudio-standards/standards/Standards.ScheduleRuleset.rb
```
to the `OpenstudioStandards::Schedules` module

### Pull Request Author
 - [x] If methods have been deprecated, update rest of code to use the new methods

### Review Checklist

 - [x] Check rubocop errors
 - [x] Check yard doc errors
 - [x] If a new feature, test the new feature and try creative ways to break it


mdahlhausen:
@weilixu I'd appreciate your review of the changes in this PR that impact PRM methods.

In particular, there are several PRM methods that get hourly values.
The hourly values methods previously returned an array of 8760 hours + 24 hours for leap years + 24 holiday values.
As far as I could tell, the holiday values were never used, so the new methods don't include that.

I also made a change to how hourly values are calculated, which is in the change from `get_8760_values_from_schedule_ruleset()` to `OpenstudioStandards::Schedules.schedule_ruleset_get_hourly_values()`. `get_8760_values_from_schedule_ruleset()` would loop through day schedules at a fixed interval and average values in the hour. It would not return the correct average if the schedule included changes at a non-fixed intervalue. I added a new method, `schedule_day_get_hourly_values` which finds the minimum interval length in a schedule day object, and uses that to do the time sampling/averaging. This gives accurately calculates the average.

lib/openstudio-standards/schedules/information.rb
```diff
@@ -40,14 +374,232 @@ def self.schedule_ruleset_get_min_max(schedule_ruleset)
...
+    #
+    # @param schedule_ruleset [OpenStudio::Model::ScheduleRuleset] OpenStudio ScheduleRuleset object
+    # @return [Array<Double>] Array of hourly values for the year
+    def self.schedule_ruleset_get_hourly_values(schedule_ruleset)
+      # define the start and end date
+      year_start_date = nil
+      year_end_date = nil
+      if schedule_ruleset.model.yearDescription.is_initialized
+        year_description = schedule_ruleset.model.yearDescription.get
+        year = year_description.assumedYear
```
weilixu, line 502: We actually have an internal discussion on the assumedYear. Does this assumedYear guaranteed a non-leap year?
The reason is the PRM guideline requires PRM run in a non-leap year, so we may need some functions to make sure we are not running in leap year simulation.

lib/openstudio-standards/schedules/information.rb
```diff
@@ -40,14 +374,232 @@ def self.schedule_ruleset_get_min_max(schedule_ruleset)
...
+      return annual_flh
+    end
+
+    # Returns an array of average hourly values from a ScheduleRuleset object
+    # Returns 8760 values, 8784 for leap years.
+    #
+    # @param schedule_ruleset [OpenStudio::Model::ScheduleRuleset] OpenStudio ScheduleRuleset object
+    # @return [Array<Double>] Array of hourly values for the year
+    def self.schedule_ruleset_get_hourly_values(schedule_ruleset)
+      # define the start and end date
```
weilixu, line 497: Since this is becoming a utility function, I wonder if we should add a safe guard here to make sure the argument provided is a ruleset schedule?

lib/openstudio-standards/schedules/information.rb
```diff
@@ -4,12 +4,346 @@ module OpenstudioStandards
...
+      time_interval_min = 15.0
+      previous_time_decimal = 0.0
+      times.each_with_index do |time, i|
+        time_decimal = (time.days * 24.0 * 60.0) + (time.hours * 60.0) + time.minutes + (time.seconds / 60)
+        interval_min = time_decimal - previous_time_decimal
+        time_interval_min = interval_min if interval_min < time_interval_min
+        previous_time_decimal = time_decimal
+      end
+      time_interval_min = time_interval_min.round(0).to_i
+
```
weilixu, line 326: Just a few more checks here might worth considering:
`0<time_interval_min && 60%time_interval_min == 0 for even division`.

weilixu (approved):
Thank you @mdahlhausen That was a lot work to make the repo organized and improved!
I have a few comments as suggestions. I think your changes make sense and do not see any major impact to PRM routine.

lib/openstudio-standards/schedules/information.rb
mdahlhausen: can't force an even division, because the average won't be correct in some cases with non-factors of 60. Consider a time interval of 7 minutes - can't use a lower number without returning the wrong average.
```
(7..60).step(7).to_a
=> [7, 14, 21, 28, 35, 42, 49, 56]
```

The time_interval_min being > zero is enforced by times being sequential in the schedule_day object.

lib/openstudio-standards/schedules/information.rb
mdahlhausen: added guard clauses to schedule_ruleset methods

lib/openstudio-standards/schedules/information.rb
mdahlhausen: @weilixu

It does not guarantee a non-leap year. How about at the start of the PRM methods, adding a check for the year, and setting to a default non-leap year if it is a leap year?

mdahlhausen: @weilixu looking more at the PRM code, there are a few places that use `schedule_get_hourly_values`. It seems those places specifically loop through weeks and hours to get exactly 8760 values, [get_fan_hours_per_week](https://github.com/NREL/openstudio-standards/blob/c1fbb2ff55b05b49599d892472e47c615c37e504/test/90_1_prm/prm_test_helper.rb#L2247) for example:
```
  def get_fan_hours_per_week(model, air_loop)
    fan_schedule = air_loop.availabilitySchedule
    fan_hours_8760 = OpenstudioStandards::Schedules.schedule_get_hourly_values(fan_schedule)
    fan_hours_52 = []

    hr_of_yr = -1
    (0..51).each do |iweek|
      week_sum = 0
      (0..167).each do |hr_of_wk|
        hr_of_yr += 1
        week_sum += fan_hours_8760[hr_of_yr]
      end
      fan_hours_52 << week_sum
    end
    max_fan_hours = fan_hours_52.max
    return max_fan_hours
  end

I'm wondering if we need to make any changes if the PRM methods that use the methods all force 8760 values anyways.
```

mdahlhausen: to be safe, it seems the best way forward is to fix the year to a non-leap year if the AssumedYear is a leap year.

mdahlhausen: @weilixu https://github.com/NREL/openstudio-standards/pull/1660

PR #1731, opened 2024-04-15
Split out residential ventilation methods

mdahlhausen:
Splits out model_add_residential_erv into model_add_residential_erv and model_add_residential_ventilator. The method was previously adding ventilation with no heat recovery, and the zero values were causing issues in an EnergyPlus method. The split adds a ventilator instead when energy recovery is not required, along with a corresponding exhaust fan.

@lymereJ would appreciate a look over this change. Is there a reason to keep the residential erv implementation as it was and come up with another work around.

If it's ok, I'll update the performance/regression tests. The performances values are quite different now.

Pull request overview
---------------------

This fixes an issue in earlier versions of 90.1, where there isn't an energy recovery requirement, so the method would add an [energy recovery ventilator with zero energy recovery](https://github.com/NREL/openstudio-standards/blob/3b02a398a1d2f9853227347c431a76390e25ff93/lib/openstudio-standards/prototypes/common/objects/Prototype.HeatExchangerAirToAirSensibleAndLatent.rb#L104-L114)
`enthalpy_recovery_ratio` is nil, because there is [no 'energy_recovery' data for 90.1-2007:](https://github.com/NREL/openstudio-standards/blob/3b02a398a1d2f9853227347c431a76390e25ff93/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb#L7360-L7378)

Thanks to @eringold for point this out.

### Pull Request Author
 - [ ] Added tests for added methods
 - [ ] If the code adds new `require` statements, ensure these are in core ruby or add to the gemspec

### Review Checklist

 - [ ] Check rubocop errors
 - [ ] Check yard doc errors


lymereJ:
@mdahlhausen - I'm just starting to look at that now even though it was merged. Could you provide details on why/how were the results different? I'm a little bit surprised to see performance results differences if all the propose code does is not model an heat exchanger if there is not energy recovery requirement.

mdahlhausen:
@lymereJ; sorry for the fast merge - needed a release candidate out for the OpenStudio build.
There results were different because of differences in fan energy use. I replaced the ERV with both a dedicate ventilation fan and dedicated exhaust fan. Perhaps the fan properties weren't aligned. It may make sense to make the supply fan near zero energy, and shift that on to the exhaust fan, to better reflect actual operation.

lymereJ:
@mdahlhausen - Thanks! Would it be alright if I create an issue to come back to this later? I would like to keep us focused on the database update for the next few months.

mdahlhausen:
@lymereJ please do!
To be clear, this only impacts the highrise and midrise apartment models. We don't use those in any projects anymore, instead deferring to HPXML models which have far better accuracy for residential modeling.

PR #1760, opened 2024-06-12
Add dedicated secondary loop for 90.1 PRM plant loops

jugonzal07:
Pull request overview
---------------------

This pull implements dedicated secondary loops for the ASHRAE 90.1 PRM plant loops that utilize constant primary, variable secondary configurations. This required updating the hardcoded value that skipped all the secondary loop logic, correctly implementing staging using EMS for the inlet pumps to each chiller, and updating the operation scheme to match the new logic.  Allows for adding up to 3 chillers with logic in place to catch scenarios when more than 3 are added to a loop to consolidate them.

### Pull Request Author

### Review Checklist

 - [ ] Check rubocop errors
 - [ ] Check yard doc errors


lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -91,32 +91,70 @@ def plant_loop_apply_prm_number_of_cooling_towers(plant_loop)
...
   # Splits the single chiller used for the initial sizing run
-  # into multiple separate chillers based on Appendix G.
-  #
+  # into multiple separate chillers based on Appendix G. Also applies
+  # EMS to stage chillers properly
   # @param plant_loop [OpenStudio::Model::PlantLoop] chilled water loop
-  # @param sizing_run_dir [String] sizing run directory
   # @return [Boolean] returns true if successful, false if not
-  def plant_loop_apply_prm_number_of_chillers(plant_loop, sizing_run_dir = nil)
+  def plant_loop_apply_prm_number_of_chillers(model, plant_loop)
```
lymereJ, line 98: I commenting here because I'm seeing this change but I think that it applies to other methods in that same file.

I don't think we need `model` as an argument because this file is about plant loop methods. Also, `model` can be obtained if needed by `plant_loop.model`.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -135,114 +173,367 @@ def plant_loop_apply_prm_number_of_chillers(plant_loop, sizing_run_dir = nil)
...
-    plant_loop.supplyComponents.each do |sc|
-      if sc.to_ChillerElectricEIR.is_initialized
-        chillers << sc.to_ChillerElectricEIR.get
-      elsif sc.to_PumpConstantSpeed.is_initialized
-        pumps << sc.to_PumpConstantSpeed.get
-      elsif sc.to_PumpVariableSpeed.is_initialized
-        pumps << sc.to_PumpVariableSpeed.get
+    if chillers.length > num_chillers
+      OpenStudio.logFree(OpenStudio::Error, 'prm.log', "For #{plant_loop.name}, the existing number of chillers exceeds the recommended amount. We have not accounted for this in the codebase yet.")
+    end
```
lymereJ, line 178: I'm not sure if understand why this is needed. Could you please clarify?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -135,114 +173,367 @@ def plant_loop_apply_prm_number_of_chillers(plant_loop, sizing_run_dir = nil)
...
+
+    # Determine the per-chiller capacity and sizing factor
+    per_chiller_sizing_factor = (1.0 / num_chillers).round(2)
```
lymereJ, line 181: Shouldn't this be 1 since the chiller are hard sized using `per_chiller_cap_w`?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -135,114 +173,367 @@ def plant_loop_apply_prm_number_of_chillers(plant_loop, sizing_run_dir = nil)
...
+
+      chiller.setName("#{template} #{chiller_cooling_type} #{chiller_compressor_type} Chiller #{i + 1} of #{num_chillers}")
+      chiller.setSizingFactor(per_chiller_sizing_factor)
+      chiller.setReferenceCapacity(per_chiller_cap_w)
+      chiller.setCondenserType(chiller_cooling_type)
+      chiller.additionalProperties.setFeature('compressor_type', chiller_compressor_type)
+
+      # Add inlet pump
+      new_pump = OpenStudio::Model::PumpVariableSpeed.new(plant_loop.model)
+      new_pump.setName("#{chiller.name} Inlet Pump")
```
lymereJ, line 215: We should probably call it `"#{chiller.name} Primary Pump"`

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -135,114 +173,367 @@ def plant_loop_apply_prm_number_of_chillers(plant_loop, sizing_run_dir = nil)
...
       end
+
+      chiller.setName("#{template} #{chiller_cooling_type} #{chiller_compressor_type} Chiller #{i + 1} of #{num_chillers}")
+      chiller.setSizingFactor(per_chiller_sizing_factor)
+      chiller.setReferenceCapacity(per_chiller_cap_w)
+      chiller.setCondenserType(chiller_cooling_type)
+      chiller.additionalProperties.setFeature('compressor_type', chiller_compressor_type)
+
+      # Add inlet pump
+      new_pump = OpenStudio::Model::PumpVariableSpeed.new(plant_loop.model)
```
lymereJ, line 214: Why do we have to use a variable speed pumps? These should be constant speed pumps.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -135,114 +173,367 @@ def plant_loop_apply_prm_number_of_chillers(plant_loop, sizing_run_dir = nil)
...
+      new_pump.setName("#{chiller.name} Inlet Pump")
+      new_pump.setRatedPumpHead(original_pump.ratedPumpHead / num_chillers)
+      new_pump.setCoefficient1ofthePartLoadPerformanceCurve(original_pump.coefficient1ofthePartLoadPerformanceCurve)
```
lymereJ, line 217: If we need to use variable speed pump, since the pump should be a constant speed pump, we should either 1) use the coefficient of a pump "riding" the curves (they're already used in the code base) or 2) have all coefficients be 0 except the constant one to not model and part load improvement.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -91,32 +91,70 @@ def plant_loop_apply_prm_number_of_cooling_towers(plant_loop)
...
+        pumps << sc.to_PumpVariableSpeed.get
+      end
+    end
+
+    # Get existing plant loop pump. We'll copy this pumps parameters before removing it. Throw exception for multiple pumps on supply side
+    if pumps.size.zero?
+      OpenStudio.logFree(OpenStudio::Error, 'prm.log', "For #{plant_loop.name}, found #{pumps.size} pumps. A loop must have at least one pump.")
+      return false
+    elsif pumps.size > 1
+      OpenStudio.logFree(OpenStudio::Error, 'prm.log', "For #{plant_loop.name}, found #{pumps.size} pumps, cannot split up per performance rating method baseline requirements.")
```
lymereJ, line 141: What does `split up per performance rating method` refers to here?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -135,114 +173,367 @@ def plant_loop_apply_prm_number_of_chillers(plant_loop, sizing_run_dir = nil)
...
+    has_secondary_plant_loop = !plant_loop.demandComponents(OpenStudio::Model::HeatExchangerFluidToFluid.iddObjectType).empty?
+
+    if has_secondary_plant_loop
+      # Add EMS to stage chillers if there's a primary/secondary configuration
+      if num_chillers > 3
+        OpenStudio.logFree(OpenStudio::Error, 'prm.log', "For #{plant_loop.name} has more than 3 chillers. We do not have an EMS strategy for that yet.")
+      elsif num_chillers > 1
+        add_ems_for_multiple_chiller_pumps_w_secondary_plant(model, plant_loop)
+      else
+        OpenStudio.logFree(OpenStudio::Info, 'openstudio.standards.PlantLoop', "No EMS for multiple chillers required for  #{plant_loop.name}, as there's only 1 chiller.")
```
lymereJ, line 241: I'm not sure if users would understand what the EMS for multiple chillers is here. Actually, I wonder if we need to log a message at all, or perhaps we should use `OpenStudio::Debug`.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -135,114 +173,367 @@ def plant_loop_apply_prm_number_of_chillers(plant_loop, sizing_run_dir = nil)
...
-      first_chiller = chillers[0]
+    OpenStudio.logFree(OpenStudio::Info, 'openstudio.standards.PlantLoop', "For #{plant_loop.name}, there are #{chillers.size} #{chiller_cooling_type} #{chiller_compressor_type} chillers.")
+
+    # Check for a heat exchanger fluid to fluid-- that lets you know if this is a primary loop
+    has_secondary_plant_loop = !plant_loop.demandComponents(OpenStudio::Model::HeatExchangerFluidToFluid.iddObjectType).empty?
+
+    if has_secondary_plant_loop
+      # Add EMS to stage chillers if there's a primary/secondary configuration
+      if num_chillers > 3
+        OpenStudio.logFree(OpenStudio::Error, 'prm.log', "For #{plant_loop.name} has more than 3 chillers. We do not have an EMS strategy for that yet.")
```
lymereJ, line 237: We probably want a error message that users could understand, I don't think that they would know what the `EMS strategy would be`.

lymereJ (commented):
Thanks for that! I think that it overall looks good. I'm leaving a few comments/questions below. Also, I think that it would be nice to include a PRM test to make sure that the correct modeling inputs are added. It would be nice to also somehow have a test to check that the pumps/chillers are correctly staged but this might a bit out of scope for now.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
lymereJ: If we do want to keep `model` as an argument, then it is missing from the YARD docs.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -135,114 +173,367 @@ def plant_loop_apply_prm_number_of_chillers(plant_loop, sizing_run_dir = nil)
...
+    ems_pump_program.addLine('SET CHILLER_PUMP_2_FLOW = CHILLER_PUMP_2_DES_FLOW,  !- A17')
+    ems_pump_program.addLine('ENDIF  !- A18')
+
+    ems_pump_program_manager = OpenStudio::Model::EnergyManagementSystemProgramCallingManager.new(model)
+    ems_pump_program_manager.setName("#{plant_name.gsub(/[-\s]+/, '_')}_Pump_Program_Manager")
+    ems_pump_program_manager.setCallingPoint('InsideHVACSystemIterationLoop')
+    ems_pump_program_manager.addProgram(ems_pump_program)
+  end
+
+  def add_ems_program_for_3_pump_chiller_plant(model, sorted_chiller_list, primary_plant)
```
lymereJ, line 473: Missing YARD docs.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -445,7 +445,7 @@ def model_create_prm_any_baseline_building(user_model, building_type, climate_zo
         next if plant_loop_swh_loop?(plant_loop)
 
         plant_loop_apply_prm_number_of_boilers(plant_loop)
-        plant_loop_apply_prm_number_of_chillers(plant_loop, sizing_run_dir)
+        plant_loop_apply_prm_number_of_chillers(model, plant_loop)
```
mdahlhausen, line 448: The CI tests are failing because the `plant_loop_apply_prm_number_of_chillers` method is expecting (OpenStudio::Model::PlantLoop, String) has input arguments, not (OpenStudio::Model::Model, OpenStudio::Model::PlantLoop)

lib/openstudio-standards/standards/Standards.PlantLoop.rb
```diff
@@ -55,7 +55,8 @@ def chw_sizing_control(model, chilled_water_loop, dsgn_sup_wtr_temp, dsgn_sup_wt
   # @param model [OpenStudio::Model::Model] OpenStudio model object
   # @return [String] common_pipe or heat_exchanger
   def plant_loop_set_chw_pri_sec_configuration(model)
-    pri_sec_config = 'common_pipe'
+    # pri_sec_config = 'common_pipe'
+    pri_sec_config = 'heat_exchanger'
```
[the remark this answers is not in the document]
jugonzal07: That's a fair point-- I'm not as familiar with how broadly this function is called. It looks like this function is called in Prototype.hvac_systems.rb under model_add_chw_loop and is used whenever the `pumping_type` is set to 'const_pri_var_sec'. This is called when setting a few different configurations:

- District Cooling
- [Heat Pumps?
](https://github.com/NREL/openstudio-standards/blob/87091ad8c6171b8ac6c96baa176d9da06614699c/lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb#L6255)
- Water Cooled chillers

That brings up a good question of if we'd like to restrict this methodology to certain configurations. If not-- we'd want to add a little more logic here.  I'm imagine the vision was the enumeration for the pump configuration would be called in many places using primary/secondary configurations and this would carry elsewhere. Happy to remove the comment.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -375,11 +380,38 @@ def model_add_chw_loop(model,
...
+      # 1.188 kw/ton for a 150 ton AirCooled chiller
+      # 0.66 kw/ton for a 150 ton Water Cooled positive displacement chiller
+      case chiller_cooling_type
+      when 'AirCooled'
+        default_cop = kw_per_ton_to_cop(1.188)
+      when 'WaterCooled'
+        default_cop = kw_per_ton_to_cop(0.66)
+      else
+        default_cop = kw_per_ton_to_cop(0.66)
+      end
```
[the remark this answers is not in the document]
jugonzal07: This was in the original code, I only moved it near the top of the code block here to ensure it was consistently called. I kept it for the sake of ensuring the code called the same efficiency as before but didn't give it much more thought than that admittedly.

lib/openstudio-standards/prototypes/common/objects/Prototype.hvac_systems.rb
```diff
@@ -375,11 +380,38 @@ def model_add_chw_loop(model,
...
+
+      # Create list of consolidated chiller names. This will reduce the number of chillers to no greater than 3 chillers
+      chiller_name_list = ["#{chiller_condenser_type}_first_stage"]
+      if num_chillers > 3
+        OpenStudio.logFree(OpenStudio::Error, 'openstudio.standards.PlantLoop', "EMS Code for multiple chiller pump has not been written for greater than 3 chillers. This has #{num_of_chillers} chillers")
```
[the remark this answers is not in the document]
jugonzal07: The EMS would not always be added. As you'll see here, it's only added in cases where a chilled water loop has 2 or 3 chillers.

https://github.com/NREL/openstudio-standards/blob/cb46fd1a3bec9c94d637259d2443cc04c44f4884/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb#L325-L331

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
jugonzal07: Good point! I'll remove it.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
jugonzal07: The way I figure, the function does not explicitly prohibit the plant loop that's being passed into the function from having multiple chillers. If the required number of chillers has fewer chillers than what's already in the loop. The code only explicitly adds chillers to the loop-- it does not remove them. The loop below it:

https://github.com/NREL/openstudio-standards/blob/cb46fd1a3bec9c94d637259d2443cc04c44f4884/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb#L186-L224

Note how it only iterates over the number of required chillers, it does not remove any.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
jugonzal07: This was a decision to follow what we did in TSPR. We once used PumpConstantSpeed pumps but moved to PumpVariableSpeed with curves that immitate constant speed. I don't recall why we moved away from them, but I'm fairly certain it had to do with how the EMS performed with PumpConstantSpeed.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
jugonzal07: That's a great point and one I missed when refactoring this. I should update this.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
jugonzal07: I am unsure, that code block for `if pump.size.zero?` was in place before I refactored this and I moved it up. It looks like the last few updates to this code block were written by you and @weilixu :

https://github.com/NREL/openstudio-standards/blame/master/lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb#L168

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
jugonzal07: Good point! Honestly, happy to remove it altogether. It's not very valuable.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
jugonzal07: Sure that's fair. I'm not quite sure how to communicate that. The issue I'm trying to communicate is that the EMS code we developed to stage chillers was only written to handle up to 3 chillers. The existing logic in the code cannot handle more than 3 chillers. That should only happen in the case of a requirement of > 2400 tons.

lib/openstudio-standards/standards/Standards.Model.rb
jugonzal07: Yup! Fixing that now, thanks Matt.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
jugonzal07: Good eye, I will add that now.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
lymereJ: I see, thanks!

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
lymereJ: Should we had a note in the code to document the choice?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
```diff
@@ -91,32 +91,72 @@ def plant_loop_apply_prm_number_of_cooling_towers(plant_loop)
...
+      OpenStudio.logFree(OpenStudio::Error, 'prm.log', "For #{plant_loop.name}, found #{pumps.size} pumps. A loop must have at least one pump.")
+      return false
+    elsif pumps.size > 1
+      OpenStudio.logFree(OpenStudio::Error, 'prm.log', "For #{plant_loop.name}, found #{pumps.size} pumps, cannot split up per performance rating method baseline requirements.")
+      return false
+    else
+      original_pump = pumps[0]
+    end
+
+    return true if chillers.empty?
```
weilixu, line 149: why do we return `true` if an error? This seems to be part of the existing code but I am not sure why it is true.
Does the boolean has any meaning?

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
weilixu: I cannot recall this change I made...

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
weilixu: We can simply say we cannot handle the scenario when there are more than 3 chiller.
In addition, I am not sure if we should stop processing or force this to be 3 chillers and then let user know so that they can modify the data themselves?
@jugonzal07 @lymereJ  thoughts?

lymereJ: That sounds good to me. I'd be in favor of stopping the process so users don't end-up with a non-valid model if they don't look at the log messages.

lib/openstudio-standards/standards/Standards.PlantLoop.rb
```diff
@@ -55,7 +55,7 @@ def chw_sizing_control(model, chilled_water_loop, dsgn_sup_wtr_temp, dsgn_sup_wt
   # @param model [OpenStudio::Model::Model] OpenStudio model object
   # @return [String] common_pipe or heat_exchanger
   def plant_loop_set_chw_pri_sec_configuration(model)
-    pri_sec_config = 'common_pipe'
+    pri_sec_config = 'heat_exchanger'
```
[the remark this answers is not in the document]
jugonzal07: Yes, that is correct. This would have wider consequences. I thought this `plant_loop_set_chw_pri_sec_configuration` was a bit goofy but didn't want to break convention. Would we prefer to limit this to a subset? I'm not sure on the intended scope of this update.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
jugonzal07: I'm unsure-- I followed the convention from earlier in the function call to maintain consistency. It appears that this function returns `true` if any part of the plant loop suggests it does not apply:

https://github.com/NREL/openstudio-standards/compare/master...primary_secondary_loops#diff-5735b63aaf6176020260a80f43cc69fa1ffc4c9a4f5e44f4524fd70b69ee2100R99-R102

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
jugonzal07: Right-- I believe it's better to stop the process. I can update the language on it to be more generic if you'd like, and not reference the EMS.

weilixu:
@mdahlhausen This PR is completed and passed the CI. Could you do a review for its merge? Once approved, we will resolve the conflict.

jugonzal07:
@mdahlhausen -- I went ahead and fixed the merge conflicts, they were related to rubocop fixes as far as I could tell and not in any way relevant to the PR updates.

mdahlhausen (approved):

mdahlhausen:
@jugonzal07 Looks good to me. Perhaps worth going through and making sure all the conversations are resolved. There is one on renaming the pump name for example that looks unaddressed.

Merge at will.

lib/openstudio-standards/standards/Standards.PlantLoop.rb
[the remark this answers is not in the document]
lymereJ: Looks like this was changed.

lib/openstudio-standards/standards/ashrae_90_1_prm/ashrae_90_1_prm.PlantLoop.rb
lymereJ: Looks like this was resolved.

PR #1837, opened 2024-10-22
Building Energy Standards VRF Data Update

leijerry888:
Updates have been made to the Building Energy Standards Data repository, see [here](https://github.com/pnnl/building-energy-standards-data/commit/7eca1a76e67047f8282457348e1899c150f1b6c6). This pull request was automatically generated.


test/os_stds_methods/test_apply_hvac_efficiency_standard.rb
```diff
@@ -50,6 +50,28 @@ def test_pthp_90_1_2019
...
+    if std.model_run_sizing_run(model, File.join(File.dirname(__FILE__),"output/#{test_name}/SR1")) == false
+      return false
+    end
+
+    # Apply the HVAC efficiency standard
+    std.model_apply_hvac_efficiency_standard(model, 'ASHRAE 169-2013-4A')
+    model.save(File.join(File.dirname(__FILE__),"output/#{test_name}.osm"), true)
+
+    # Spot check expected efficiencies
+    vrfs = model.getAirConditionerVariableRefrigerantFlows
```
mdahlhausen, line 70: Add a `.sort` to the end of this so it passes on all machines; depending on the machine it may return in a different order. Or you could find another way to check a specific efficiency.

mdahlhausen:
Models with VRF that are 90.1-2007 or earlier will error out because there isn't a data .json for vrfs.
`model.getAirConditionerVariableRefrigerantFlows.sort.each { |obj|air_conditioner_variable_refrigerant_flow_apply_efficiency_and_curves(obj) }`

As you did with the NECB code, you'll need to add a `Standards.AirConditionerVariableRefrigerantFlow.rb` file for 90.1-2007, 90.1-2004, DOE Ref 1980-2004, and DOE Ref Pre-1980 that sets a dummy `air_conditioner_variable_refrigerant_flow_apply_efficiency_and_curves` to return null. Or, you could copy the 90.1-2010 data for those earlier templates. Either approach works.

mdahlhausen (changes requested):
See comments

lymereJ:
> Models with VRF that are 90.1-2007 or earlier will error out because there isn't a data .json for vrfs. `model.getAirConditionerVariableRefrigerantFlows.sort.each { |obj|air_conditioner_variable_refrigerant_flow_apply_efficiency_and_curves(obj) }`
> 
> As you did with the NECB code, you'll need to add a `Standards.AirConditionerVariableRefrigerantFlow.rb` file for 90.1-2007, 90.1-2004, DOE Ref 1980-2004, and DOE Ref Pre-1980 that sets a dummy `air_conditioner_variable_refrigerant_flow_apply_efficiency_and_curves` to return null. Or, you could copy the 90.1-2010 data for those earlier templates. Either approach works.

Addressed! I went with the latter to avoid adding more code that doesn't do anything, also, I think that in retrospect it might be good to offer data for earlier code version.

mdahlhausen (approved):
Added support for DOE References building standards and updated the 90.1-2019 vrf test

Will merge once CI test finishes

PR #1840, opened 2024-10-23
Building Energy Standards Economizers Data Update

leijerry888:
Updates have been made to the Building Energy Standards Data repository, see [here](https://github.com/pnnl/building-energy-standards-data/commit/bea47dcde9b478b78a4d8b1f6c46f46ca8def928). This pull request was automatically generated.


lib/openstudio-standards/prototypes/common/objects/Prototype.Model.rb
```diff
@@ -2050,14 +2053,19 @@ def apply_economizers(climate_zone, model)
         # specified, economizer requirements were aimed at comfort
         # cooling, not computer room cooling (as per input from the MSC).
 
+        # Process climate zone:
+        # Moisture regime is not needed for climate zone 8
+        climate_zone = climate_zone.split('-')[-1]
```
mdahlhausen, line 2058: I'm a little confused on the terminology here. `climate_zone_code` is the full name, and `climate_zone` becomes just the number/letter combo? climate_zone as the number/letter combo and isn't used later on. Why include this line then and make a separate `climate_zone_code`? It appears in other files too.

lymereJ: Both the "short" climate zones and the "long" climate zones are needed. 
- `climate_zone` (e.g., "1A") is needed in the `search_criteria`
- `climate_zone_code` (e.g., "ASHRAE 169-2013-1A") is needed further down, for example on line 2098 `unless air_loop_hvac_economizer_type_allowable?(air_loop, climate_zone_code)`

lymereJ: Note that this approach already exists in the code base, see [here](https://github.com/NREL/openstudio-standards/blob/master/lib/openstudio-standards/standards/Standards.ThermalZone.rb#L249-L250), although it looks like the meaning of these two variables are inverted.

PR #1896, opened 2025-02-17
Building Energy Standards Exterior Lighting Data Update

leijerry888:
Updates have been made to the Building Energy Standards Data repository, see [here](https://github.com/pnnl/building-energy-standards-data/commit/73320d797a3b1972a1da15f507ee8ff3c4fa8ad6). This pull request was automatically generated.


lib/openstudio-standards/prototypes/common/objects/Prototype.Model.exterior_lights.rb
```diff
@@ -47,37 +67,61 @@ def model_add_typical_exterior_lights(model, exterior_lighting_zone_number, onsi
...
-    if !exterior_lighting_properties['occupancy_setback_reduction'].nil? && (exterior_lighting_properties['occupancy_setback_reduction'] > 0.0)
+
+    occupancy_setback_reduction = 0.0
+    ext_lights_sch_other = model.alwaysOnDiscreteSchedule
+    if exterior_lighting_properties.has_key?('occupancy_setback_reduction')
+      if !exterior_lighting_properties['occupancy_setback_reduction'].nil? && (exterior_lighting_properties['occupancy_setback_reduction'] > 0.0)
+        occupancy_setback_reduction = exterior_lighting_properties['occupancy_setback_reduction']
+      end
+    else
+      unless ['90.1-2004', '90.1-2007', 'DOE Ref Pre-1980', 'DOE Ref 1980-2004', 'DEER 1985', 'DEER 1996', 'DEER 2003', 'DEER 2007', 'DEER Pre-1975'].include?(template)
```
mdahlhausen, line 78: make an inclusive criteria, not an exclusive one. There are several other standards ignored here.

lib/openstudio-standards/prototypes/common/objects/Prototype.Model.exterior_lights.rb
```diff
@@ -47,37 +67,61 @@ def model_add_typical_exterior_lights(model, exterior_lighting_zone_number, onsi
...
+        power = exterior_lighting_properties['parking_areas_and_drives']
+      else
+        search_criteria = {
+          'template' => template,
+          'lighting_zone' => exterior_lighting_zone_number,
+          'allowance_type' => 'parking lots and drives'
+        }
+        exterior_lighting_parking_req = standards_lookup_table_first(table_name: 'exterior_lighting', search_criteria: search_criteria)
+        power = exterior_lighting_parking_req['allowance']
+        if exterior_lighting_parking_req['allowance_unit'] == 'W'
```
mdahlhausen, line 107: in what instance is the allowance in watts?

mdahlhausen (changes requested):
I have another major exterior lighting module change I'm working on. Address the minor comments, and I'll see if it makes more sense to merge this PR into mine or vice-versa.

lib/openstudio-standards/prototypes/common/objects/Prototype.Model.exterior_lights.rb
lymereJ: One example is the base allowance.

lib/openstudio-standards/prototypes/common/objects/Prototype.Model.exterior_lights.rb
lymereJ: Which ones are missing? I can certainly do that but there are a lot of applicable ones as well.

mdahlhausen: other DEER versions, the NZE templates. I'm wondering if this should go back to being something in the standard fields? If it toggles on the standard, it should probably go in there.

lib/openstudio-standards/prototypes/common/objects/Prototype.Model.exterior_lights.rb
mdahlhausen: Makes sense for base allowance. It seems like this switch statement applies to all lighting types. A lot of duplicate code. It's not necessary in parking and most others, because all standard entries are in W/ft2. I can handle the cleanup unless there is a way you'd prefer to do it.

lymereJ: I removed the statements and also clarified the names of the definitions to be more in line with the inputs.

lymereJ:
Thanks!

- I was hoping that we could use the full name instead of the abbreviated one. It's easier to export the data if the file has the same name as the table name but that's okay, we'll work around that in the database repo.
- Thanks for catching that, I'll make the change in the database.
- Yes, I wasn't sure how you wanted to handle this. I left it as-is since it produced the same results as before. I'll add handling for the W/ft case then.

mdahlhausen:
> Thanks!
> 
>     * I was hoping that we could use the full name instead of the abbreviated one. It's easier to export the data if the file has the same name as the table name but that's okay, we'll work around that in the database repo.
> 
>     * Thanks for catching that, I'll make the change in the database.
> 
>     * Yes, I wasn't sure how you wanted to handle this. I left it as-is since it produced the same results as before. I'll add handling for the W/ft case then.

I suppose we can leave the longer name - I'll check that it works on the windows environments with the most extreme path lengths

For the building facades, perhaps we just default to using the W/ft2 path?

lymereJ:
That's fine, I've made the changes locally, I'll push them today. Sounds good for the default. I'll push it also.

Issue #246, opened 2017-05-23
Cloning of construction types doesn't happen


jmarrec:
https://github.com/NREL/openstudio-standards/blob/master/openstudio-standards/lib/openstudio-standards/standards/Standards.Model.rb#L3419

It doesn't clone it.

If you take a construction set from the OS Library, for eg 90.1-2010 MidriseApartment, the same construction is applied to the ground contract walls, ceiling and floors (and ExteriorFloor)
After applying `model.apply_prm_construction_types('90.1-2013')`, it just changed the existing one with the latest in the loop (GroundContactWall is last added, [#L3500](https://github.com/NREL/openstudio-standards/blob/master/openstudio-standards/lib/openstudio-standards/standards/Standards.Model.rb#L3500)), and not cloning happened

```
 OS:StandardsInformation:Construction,
   {f84c35d2-f1ea-430e-86a5-40f829e6bf82}, !- Handle
   {ca0b8402-a973-458c-a138-eacbdf3dea5d}, !- Construction Name
-  ExteriorFloor,                          !- Intended Surface Type
-  ,                                       !- Standards Construction Type
+  GroundContactWall,                      !- Intended Surface Type
+  Mass,                                   !- Standards Construction Type
```

mdahlhausen:
Going through old bugs and making notes:

This method deals with assigned standard constructions to surfaces in the PRM. Seems like a fix for PNNL @weilixu @lymereJ .

The method `model_apply_prm_construction_types` sets standards information for constructions, in case they were not set before.
The method `model_apply_standard_constructions` goes through each construction for a list of types and assigns the standard values to it.
The issue is that some constructions apply to multiple kinds of surfaces in the model. In this case, the same construction will get edited, and ultimately end up with the values for the last surface type / construction type pair in the loop, regardless if it is appropriate for all surface types. At the extreme, imagine a model with the same construction for all surfaces. In that case, the method will set construction properties for that construction to the last surface type / construction pair, which will impact all surfaces in the model.

The fix is to clone existing constructions to create a new standards construction for each surface type / construction type pair, and then apply it to those constructions.

It would also be nice to use the 90.1-2019 PRM methods as the default`model_apply_prm_construction_types` and `model_apply_standard_constructions`, adding extra arguments if necessary.

weilixu:
@mdahlhausen 
We are looking this issue now.
[...]

PR #1903, opened 2025-02-26
Use prm apply construction method in standards

weilixu:
Pull request overview
---------------------

Move the model apply constructions to Standard model. 
Some updates are made to the module functions that requires review.

### Pull Request Author

 - [ ] Method changes or additions
 - [ ] Added tests for added methods
 - [ ] Documented new methods using [yard syntax](https://rubydoc.info/gems/yard/file/docs/GettingStarted.md)
 - [ ] Resolved yard documentation errors for new code (ran `bundle exec rake doc`)
 - [ ] Resolved rubocop syntax errors for new code (ran `bundle exec rake rubocop`)
 - [ ] If the code adds new `require` statements, ensure these are in core ruby or add to the gemspec

### Review Checklist

 - [ ] Check rubocop errors
 - [ ] Check yard doc errors


weilixu:
This PR should be draft for review.
@mdahlhausen one thing I changed is to lower the log level in the construction_calculated_fenestration_u_factor function when SQL is not found. Currently, the error message fails all non prm test cases.
The reason is because PRM has a function called space_conditioning_category that requires the model being simulated. The non prm test cases do not perform simulation so the SQL files are missing when calling this function.
I updated the space_conditioning_category to run without the SQL file.

weilixu:
Follow up on another issues found from CI results.

`bldg_11` has windows that failed the function -> subsurface.uFactor. I suspect this was due to the windows were using layer by layer method.
`bldg_17` i am not entirely sure why the space `3.Apartment 13` is categorized as UNCONDITIONED. It is perhaps due to the missing design loads of the space.

I added sizing run for both cases and the test passed.
The downside with this walk-around is that the additional sizing run could make CI run longer.
The positive side is there are only 5 cases need additional sizing run:
Progress :: |[31mFAILED: test_bldg_17_midrise_apartment_lowrise_iecc_constructions IN FILE test/90_1_prm/test_prm_baseline_bldg.rb[0m
=============[31mFAILED: test_bldg_11_num_boilers IN FILE test/90_1_prm/test_prm_baseline_bldg11.rb[0m
=============[31mFAILED: test_bldg_11_num_chillers IN FILE test/90_1_prm/test_prm_baseline_bldg11.rb[0m
=============[31mFAILED: test_bldg_11_plant_controls IN FILE test/90_1_prm/test_prm_baseline_bldg11.rb[0m
=============[31mFAILED: test_bldg_11_shw IN FILE test/90_1_prm/test_prm_baseline_bldg11.rb[0m
==============|

mdahlhausen:
Changes look fine but I'm getting some failed simulations running the bldg11 tests. Lots of errors like:
`Detailed="SUB SURFACE 155" has an opaque surface construction; it should have a window construction.`
Looking into it...

mdahlhausen (changes requested):
There is an error with the 'Any Vertical Glazing' standards_construction_type key, as it only exists in the 90.1-2019 PRM template. See comment and fix.

The other changes look great. glad it's consolidating the method. The 'Warn' instead of 'Error' in the SQL lookup is fine with me.

weilixu:
> There is an error with the 'Any Vertical Glazing' standards_construction_type key, as it only exists in the 90.1-2019 PRM template. See comment and fix.
> 
> The other changes look great. glad it's consolidating the method. The 'Warn' instead of 'Error' in the SQL lookup is fine with me.

Thanks for checking. I will look into the error you mentioned here.

lib/openstudio-standards/standards/Standards.Model.rb
```diff
@@ -4129,9 +4146,38 @@ def model_apply_prm_construction_types(model)
...
+    types_to_modify << ['GroundSlabPreprocessorAverage', 'GroundContactFloor', 'Unheated']
+    types_to_modify << ['GroundSlabPreprocessorCore', 'GroundContactFloor', 'Unheated']
+    types_to_modify << ['GroundSlabPreprocessorPerimeter', 'GroundContactFloor', 'Unheated']
+
+    # Basement preprocessor
+    types_to_modify << ['GroundBasementPreprocessorAverageWall', 'GroundContactWall', 'Mass']
+    types_to_modify << ['GroundBasementPreprocessorAverageFloor', 'GroundContactFloor', 'Unheated']
+    types_to_modify << ['GroundBasementPreprocessorUpperWall', 'GroundContactWall', 'Mass']
+    types_to_modify << ['GroundBasementPreprocessorLowerWall', 'GroundContactWall', 'Mass']
+
```
mdahlhausen, line 4180: Adding all of these options further exposed a problem described in #246:
```
  # Modify all constructions of each type
  types_to_modify.each do |boundary_cond, surf_type, const_type|
    constructions = OpenstudioStandards::Constructions.model_get_constructions(model, boundary_cond, surf_type)

    constructions.sort.each do |const|
      standards_info = const.standardsInformation
      old_surf_type = standards_info.intendedSurfaceType
      old_const_type = standards_info.standardsConstructionType
      standards_info.setIntendedSurfaceType(surf_type)
      standards_info.setStandardsConstructionType(const_type)
    end
  end
```
The modifications override the intended surface type and standards construction type for each construction object. The problem is that some constructions are used across multiple combinations of (boundary condition/surface type/construction type). The construction will end up with whatever is last in the array.

For example, an exterior floor construction may start with 'ExteriorFloor' intended surface type and no standards construction type, and end up 'GroundContactWall' as the intended surface type and 'Mass' as the construction type.

This piece of code needs to deep clone a construction before setting the intended surface type and construction type if used elsewhere.

This was caught because it the floor construction in the model ended up with 'GroundContactWall' as the intended surface type, and it was trying to calculate wall height to set properties, but failed because it was not a wall.

Fix #246 as part of this merge.

mdahlhausen (changes requested):
See issue above, related to #246.

weilixu:
The code is still messy but the changes are more significant than I originally planned.
1. A simple duplication does not work because in the `planar_surface_apply_standard_construction` the surface type (`surf_type`) is obtained from construction. If a construction's intended surface is misassigned, then the surface will generate incorrect outcome.
2. The construction_deep_copy does not supporting F factor and C factor constructions - expanded this function (unit tests is not completed)
3. Surface conditioning category - it is unavoidable to have the correct surface conditional category for all PRM models. However, many legacy code test cases do not run sizing runs so there is no surface conditioning category per surface in those test cases, thus causing errors.
I added a flag function to differentiated legacy code and PRM 2019 to avoid legacy energy code test case failures. In the meantime, fixing the duplication errors.

This needs CI run to verify all these changes won't impact other routines. I need to clean up the code once verified.

lib/openstudio-standards/constructions/create.rb
```diff
@@ -5,15 +5,31 @@ module Constructions
     # Methods to create Constructions
 
     # This will create a deep copy of the construction, meaning it will clone and create new material objects as well
+    # This function will clone
     #
     # @param construction [OpenStudio::Model::Construction] OpenStudio Construction object object
     # @return [OpenStudio::Model::Construction] New OpenStudio Construction object
     def self.construction_deep_copy(construction)
```
weilixu, line 12: @mdahlhausen I modified this function to include deep copy for f factor ground floor construction, c factor underground wall construction.
I wonder if you prefer to keep a strict typing which means the return is limited to one type? If so, I can create separate deep clone functions.
Also, `OS_Construction_AirBoundary` is not included in this function, do you think it is necessary to include it in here too?

mdahlhausen: A more general deep copy function is appropriate. Thanks.

I'm getting some new errors line 26.

weilixu: > A more general deep copy function is appropriate. Thanks.
> 
> I'm getting some new errors line 26.

@mdahlhausen that is because the AirBoundary construction is not included in this function. I can include the AirBoundary type to here as well.

weilixu:
it seems in bldg 12 model, there are many tiny ground floor surfaces in some spaces that causes Eplus Ffactor construction calculation failure.
An exception is added to the function `construction_set_surface_slab_f_factor` in the Constructions module, modify.rb.

weilixu:
@mdahlhausen CI passed. The code is ready for review.

mdahlhausen (approved):
@weilixu Looks good! Thanks for fixing that bug. Clear to merge.

PR #1937, opened 2025-05-14
Disable sizing run design day simulations

mdahlhausen:
Don't run simulation for sizing periods during sizing run. It will still run zone, system, and plant sizing, but not simulate the design days.

Pull request overview
---------------------
Goal is to reduce runtime in create_typical.

### Pull Request Author

### Review Checklist


mdahlhausen:
@khaddad I'm running into a regression test error in this test:
test\necb\regression_tests\tests\test_necb_ecm_hs14_cgshp_fancoils.rb

Could you edit the [set_ghx_loop_district_cap](https://github.com/NREL/openstudio-standards/blame/4c2b381058050dfef65dc973fefa6ac65748696a/lib/openstudio-standards/standards/necb/ECMS/hvac_systems.rb#L2003) method to pull sizing from the system sizing instead?

Background:
This PR turns off running the simulation during design days in the sizing run in an effort to save simulation time on numerous model runs. The [set_ghx_loop_district_cap](https://github.com/NREL/openstudio-standards/blame/4c2b381058050dfef65dc973fefa6ac65748696a/lib/openstudio-standards/standards/necb/ECMS/hvac_systems.rb#L2003) method relies on the output .sql to be populated, as it pulls the top values from the heating or cooling rate over the design days. There are other plant sizes available that could be used instead, and as these are district heating/cooling objects, there is no performance curve so the capacity does not matter to system performance.

khaddad:
> @khaddad I'm running into a regression test error in this test: test\necb\regression_tests\tests\test_necb_ecm_hs14_cgshp_fancoils.rb
> 
> Could you edit the [set_ghx_loop_district_cap](https://github.com/NREL/openstudio-standards/blame/4c2b381058050dfef65dc973fefa6ac65748696a/lib/openstudio-standards/standards/necb/ECMS/hvac_systems.rb#L2003) method to pull sizing from the system sizing instead?
> 
> Background: This PR turns off running the simulation during design days in the sizing run in an effort to save simulation time on numerous model runs. The [set_ghx_loop_district_cap](https://github.com/NREL/openstudio-standards/blame/4c2b381058050dfef65dc973fefa6ac65748696a/lib/openstudio-standards/standards/necb/ECMS/hvac_systems.rb#L2003) method relies on the output .sql to be populated, as it pulls the top values from the heating or cooling rate over the design days. There are other plant sizes available that could be used instead, and as these are district heating/cooling objects, there is no performance curve so the capacity does not matter to system performance.

@mdahlhausen I checked the autosized values for the district heating and cooling objects (autosizedNominalCapacity) using a test case and these are equal. They seem to be set to the sum of the autosized capacities of the district heating and cooling objects. This is why I use the max capacities from the design day output variables for 'District Cooling Water Rate' and 'District Heating Water Rate'. We use these capacities to cost the ground-heat exchanger. I'll see if we can use a work-around of some sort.

